PR #41, opened 2015-04-17
Depend on the JRE. Updated RTI packages

esteve:
This adds the latest RTI packages and installs the default JRE

@dirk-thomas @tfoote @wjwwood


dirk-thomas:
lgtm

wjwwood:
+1

PR #45, opened 2015-04-22
add groovy build step to set a useful description

wjwwood:
This is the first build step to it is set shortly after the ros2/ros2 git repo is checked out. Example:

screen shot 2015-04-21 at 5 37 35 pm

For review.


tfoote:
+1

dirk-thomas:
LGTM

PR #50, opened 2015-05-12
Added new connext packages

esteve:
This adds the new connext packages that include the request/reply API.

Builds pass:

http://54.183.26.131:8080/view/ros2/job/ros2_batch_ci_linux/112/

Connects to #49

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -33,11 +33,9 @@ RUN apt-get update && apt-get install -y libopensplice64
 RUN apt-get update && apt-get install -y default-jre-headless
 
 # Add and install the RTI binaries we made.
-ADD rticonnextdds-messaging/librticonnextdds51_5.1.0-4_amd64.deb /tmp/librticonnextdds51_5.1.0-4_amd64.deb
-ADD rticonnextdds-messaging/librticonnextdds51-dev_5.1.0-4_amd64.deb /tmp/librticonnextdds51-dev_5.1.0-4_amd64.deb
-ADD rticonnextdds-messaging/rticonnextdds-tools_5.1.0-4_amd64.deb /tmp/rticonnextdds-tools_5.1.0-4_amd64.deb
-ADD rticonnextdds-messaging/librticonnextdds-messaging51_5.1.0-3_amd64.deb /tmp/librticonnextdds-messaging51_5.1.0-3_amd64.deb
-ADD rticonnextdds-messaging/librticonnextdds-messaging51-dev_5.1.0-3_amd64.deb /tmp/librticonnextdds-messaging51-dev_5.1.0-3_amd64.deb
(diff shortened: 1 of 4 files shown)


esteve:
@dirk-thomas @tfoote @wjwwood for review.

I submitted the changes to RTI's repo in https://github.com/rticommunity/rticonnextdds-src/pull/7

wjwwood:
+1

linux_docker_resources/Dockerfile
```diff
@@ -33,11 +33,9 @@ RUN apt-get update && apt-get install -y libopensplice64
...
 
 # Add and install the RTI binaries we made.
-ADD rticonnextdds-messaging/librticonnextdds51_5.1.0-4_amd64.deb /tmp/librticonnextdds51_5.1.0-4_amd64.deb
-ADD rticonnextdds-messaging/librticonnextdds51-dev_5.1.0-4_amd64.deb /tmp/librticonnextdds51-dev_5.1.0-4_amd64.deb
-ADD rticonnextdds-messaging/rticonnextdds-tools_5.1.0-4_amd64.deb /tmp/rticonnextdds-tools_5.1.0-4_amd64.deb
-ADD rticonnextdds-messaging/librticonnextdds-messaging51_5.1.0-3_amd64.deb /tmp/librticonnextdds-messaging51_5.1.0-3_amd64.deb
-ADD rticonnextdds-messaging/librticonnextdds-messaging51-dev_5.1.0-3_amd64.deb /tmp/librticonnextdds-messaging51-dev_5.1.0-3_amd64.deb
+ADD rticonnextdds-src/librticonnextdds51_5.1.0.1-1_amd64.deb /tmp/librticonnextdds51_5.1.0.1-1_amd64.deb
+ADD rticonnextdds-src/librticonnextdds51-dev_5.1.0.1-1_amd64.deb /tmp/librticonnextdds51-dev_5.1.0.1-1_amd64.deb
+ADD rticonnextdds-src/rticonnextdds-tools_5.1.0.1-1_amd64.deb /tmp/rticonnextdds-tools_5.1.0.1-1_amd64.deb
```
dirk-thomas: Is there a specific reason why we switched to use a four-part version number? Since upstream has only three adding the debinc behind that looked reasonable to me. Afaik there is no version number 5.1.0.1 upstream?

esteve: No, there's no 5.1.0.1 version upstream, but RTI changed master without tagging it, so we can't consistently create Debian packages if the upstream sources are not properly tagged. Given the case, I opted to release these packages with a new subversion.

esteve: In any case, RTI merged the Debian stuff (with the four-part version number), so no matter what, these packages are generated from the current upstream rticommunity/rticonnextdds-src repository as it is.

dirk-thomas:
Beside my comment about the version number +1.

May be we should try not to add a forth version part for the upcoming 5.2 release.

PR #52, opened 2015-06-02
added use of compiler warnings plugin in jenkins

wjwwood:
This adds entries in the job configs to parse compiler warnings out of the job console output. On Linux it parses GCC 4 and on OS X it parses clang. There is no support for Windows.

@dirk-thomas If this works out I need to add this plugin to the list for setting up the build farm. Where is that list?

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -192,6 +192,9 @@ python -u run_ros2_batch.py %CI_ARGS%""",
             'git_submodule_extension': '',
             # Do not use the ssh-agent build wrapper on Windows since ssh-agent doesn't work.
             'ssh_agent_build_wrapper': '',
+            'extra_publishers': compiler_warning_snippet.format(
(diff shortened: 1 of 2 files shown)


wjwwood:
You can see the trend on the project pages:

http://54.183.26.131:8080/job/ros2_batch_ci_osx/
http://54.183.26.131:8080/job/ros2_batch_ci_linux/

dirk-thomas:
LGTM

The Jenkins plugins are specified in the buildfarm_deployment repository in a puppet file: https://github.com/ros-infrastructure/buildfarm_deployment/blob/29b0cc1a110f7abeef444a3af4a497c155c30743/master/manifests/site.pp#L87

esteve:
+1

tfoote:
+1

wjwwood:
I also added an entry for Windows (I previously didn't notice the `MSBuild` option).

Issue #70, opened 2015-07-18
switch to Connext 5.2


dirk-thomas:
* [x] depends on the source being available to us
* [x] create new Debian package
* [ ] create VS 2015 package to fix runtime problem of prebuilt packages: ros2/ros2#51
  * [x] some of the tests already pass with Connext 5.2 Pro (which is built with VS 2013), hopefully the custom built will make all pass...
* [x] ~~either build from source for OS X or~~ test prebuilt binary packages on OS X
* [x] update rmw_connext repo: ros2/rmw_connext#75 ros2/rmw_connext#54
* [x] update tests to cover unbounded string and sequences: ros2/system_tests#17
* [x] update CI scripts #75

esteve:
Preliminary Debian packaging in:

https://github.com/rticommunity/rticonnextdds-src/pull/8

Debian packages in:

https://github.com/osrf/rticonnextdds-src/tree/binaries/ubuntu/trusty/amd64/5.2.0-0

dirk-thomas:
PR to update the Docker image: ros2/ros2#75

esteve:
Final Debian packaging already merged into upstream RTI repo:

https://github.com/rticommunity/rticonnextdds-src/commit/13830c921718cdbef801179d14ba8b53cc83a18a

Debian packages in:

https://github.com/osrf/rticonnextdds-src/tree/binaries/ubuntu/trusty/amd64/5.2.0-1

CI scripts:

https://github.com/ros2/ros2/pull/75

dirk-thomas:
I am closing this since the only remaining task is the custom Connext package built with Visual Studio 2015 which is tracked in a separate ticket.

PR #75, opened 2015-07-28
Update to RTI Connext 5.2.0-1

esteve:
Connects to ros2/ros2#70


dirk-thomas:
Since this works (http://ci.ros2.org/job/ros2_batch_ci_linux/185/) we should merge it.

PR #77, opened 2015-07-30
switch from JUnit to GoogleTest result type

dirk-thomas:
Currently we use XUnit with the JUnit result type which is unable to parse the result files generated by GTest:

    http://ci.ros2.org/job/ros2_batch_ci_osx/56/

I changed the OS X job to use XUnit with the GoogleTest result type which seems to work:

    http://ci.ros2.org/job/ros2_batch_ci_osx/57/


tfoote:
+1

Issue #72, opened 2015-07-21
Update to the official VS 2015 release


wjwwood:
It has been officially released:

http://blogs.msdn.com/b/somasegar/archive/2015/07/20/visual-studio-2015-and-net-4-6-available-for-download.aspx

We need to upgrade to it on the Windows CI slave and test it out.

j-rivero:
It should not be difficult. I was about to do it, but found a really minimal installation or configuration of Windows when connected through VPN to eatable. I can not see or touch more than the Jenkins Window. How can I interact with that? Is it possible to set it back to the standard Windows desktop?

wjwwood:
It's been like that for a while, I just restarted the machine, maybe that will fix the desktop.

j-rivero:
That made the trick, thanks William.

j-rivero:
One minor issue before installing. You are currently passing `amd64` as argument to vcvarsall.bat, which configures the 64 bit native compiler to generate 64 bit code. That is ideal, but for some reason, the free Express editions of VS don't support it. However, these Express versions support the `x86_amd64` code, which runs the 32 bit compiler to generate 64 bit code.

In our case, we can go with VS Professional and use the amd64 keyword, no problem. But if the building farm is designed to be reconfigurable, maybe it is better to use the Express version and/or the `x86_amd64`.

dirk-thomas:
I have create a branch for the CI scripts: https://github.com/ros2/ros2/tree/x86_amd64 which uses what you suggest (https://github.com/ros2/ros2/commit/3a0e537d575a042308fe99b3903c084fe70fc385)

You can try this by passing `x86_amd64` to the `CI_SCRIPTS_BRANCH` parameter of the job.

j-rivero:
Nice, let's try installing the Express version then.

j-rivero:
The update should be ready, to 2015 Express version. The system needed a reboot, a press the button but I can not contact with the machine again. Maybe is running Boot Camp and need to restart it on Windows? Hopefully it is not broken.

j-rivero:
The [testing run](http://ci.ros2.org/job/ros2_batch_ci_windows/107/console) since it can not find the devenv.exe utility. Our fork of [uncrustify](https://github.com/ament/uncrustify/blame/master/CMakeLists.txt) is calling it to [upgrade](https://msdn.microsoft.com/en-us/library/w15a82ay.aspx) .sln files.

[...]

PR #81, opened 2015-08-08
use x86_amd64 instead of amd64 on windows

dirk-thomas:
We are going to use the Community edition of Visual Studio 2015 now. That requires a different compiler.

Connects to ros2/ros2#72


wjwwood:
+1

tfoote:
+1

esteve:
+1

PR #87, opened 2015-08-17
rebase all repos to master which must succeed

### LINKED ISSUE #78: should the CI jobs automatically rebase to master
Currently the CI job tests the code from the specified branches. If they are not up-to-date but branches earlier and master has diverged that is not covered by the test. The result is that the CI job might succeed but after merging the branch the CI job might fail on master.

Would it be sufficient to always do this after switching to the to-be-tested branches:

    vcs custom --git --args rebase master

dirk-thomas:
Fixes #78.


wjwwood:
+1

esteve:
+1

tfoote:
If this must pass should we handle the return code?

dirk-thomas:
We don't have to. Only if passed `exit_on_error=False` the script can continue. If the command fails it will `sys.exit()` already.

tfoote:
+1

PR #96, opened 2015-08-20
set different domain ids for each os

dirk-thomas:
To prevent any kind of cross talk this patch adds different ROS_DOMAIN_IDs for each platform.

When this is merged the OpenSplice config files on the OS X and Windows slaves must be updated.

http://ci.ros2.org/job/ros2_batch_ci_linux/241/

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -28,6 +28,8 @@ RUN pip3 install -U setuptools pip virtualenv
 
 # Install the OpenSplice binary from the OSRF repositories.
 RUN apt-get update && apt-get install -y libopensplice64
+# Update default domain id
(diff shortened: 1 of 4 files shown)


esteve:
Looks good, but I'd prefer the IDs to be consecutive.

wjwwood:
lgtm

jacquelinekay:
+1

dirk-thomas:
The advantage of using numbers based on the ascii code is that I don't have to look at a mapping table to figure out who uses a specific domain id. I can easily derive the information from the number itself. Otherwise we would need to maintain that mapping table somewhere.

esteve:
I didn't know they were based on the ASCII code for each slave. Agreed, this number scheme makes more sense :-)

PR #101, opened 2015-08-28
Added support for packaging on OSX

esteve:


wjwwood:
I'm gonna cancel this job since it will hang and then fail:

http://ci.ros2.org/job/ros2_packaging_osx/1/console

You need to remove the use of ssh credentials on the OS X job. I'm not sure why, but you can look at how the other OS X jobs are currently configured (for ssh agent and git credentials). If you mimic those ones then you'll get the right behavior.

esteve:
No problem, go ahead.

esteve:
@wjwwood @dirk-thomas triggered a CI job this morning without any changes and it passed:

http://ci.ros2.org/view/packaging/job/ros2_packaging_osx/5/

I think the slave that needs to have SSH disabled is Windows, but I already removed that when I merged the Windows PR:

https://github.com/ros2/ros2/blob/ci_scripts/job_templates/ros2_packaging_job.xml.template#L136

Seeing that it seems to have generated a valid tarball, I'm going to merge this PR. @gerkey can you test the results? Thanks! http://ci.ros2.org/view/packaging/job/ros2_packaging_osx/5/artifact/workspace/ros2-package-osx.tar.bz2

dirk-thomas:
I did remove the SSH agend from the OS X job. Otherwise it did not work.

esteve:
@dirk-thomas thanks! Was that part of the job template or is it something that had to be changed in the Jenkins configuration?

gerkey:
I tried the package generated by the job and at least talker and listener work, which is great.  It doesn't have the bridge, but that's because we need to add the step of installing a bit ROS 1 first, as described in step 6 of [the from-source instructions](https://github.com/ros2/ros2/wiki/OSX-Development-Setup#install-prerequisites).

dirk-thomas:
The configuration in this PR is not equal to what the Jenkins job uses so it still needs to be updated to match.

esteve:
@dirk-thomas what are the changes needed in this PR? Could you create a separate PR with them? Thanks!

dirk-thomas:
I have fixed the OS X jobs to also work when starting with an empty workspace in 61a70e74becdc30009e8a2776594ee9f861bd824

PR #103, opened 2015-08-28
set CMAKE_BUILD_TYPE to RelWithDebInfo for packaging

dirk-thomas:
http://ci.ros2.org/view/packaging/job/ros2_packaging_linux/18/
http://ci.ros2.org/view/packaging/job/ros2_packaging_osx/6/
http://ci.ros2.org/view/packaging/job/ros2_packaging_windows/7/

The Linux and OS X jobs will likely fail on the `ros1_bridge` right now for a different reason. Windows is not using a specific build type anyway.


gerkey:
Assuming that `--cmake-args` will pass that string to CMake, +1.

PR #104, opened 2015-08-28
use platform specific folder name in archives

dirk-thomas:
Currently the archives unpack into a folder `ros2`. So if you download multiple archives and unpack them they collide.


wjwwood:
+0, I think it also makes sense if you get the same folder layout no matter what OS you download for. Doesn't matter I guess.

PR #114, opened 2015-09-08
set LANG

dirk-thomas:
Necessary to prevent unicode encoding errors:
* http://ci.ros2.org/job/ros2_batch_ci_linux/324/console
* http://ci.ros2.org/job/ros2_batch_ci_linux/326/console


wjwwood:
It's weird that this must be done manually, but I guess docker is leaner than I'm used to.

wjwwood:
+1

PR #115, opened 2015-09-08
install boost for eProsima

dirk-thomas:
This is necessary if we want to build the eProsima packages.

Connect to ros2/ros2#124

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -42,6 +42,9 @@ ADD rticonnextdds-src/librticonnextdds52_5.2.0-1_amd64.deb /tmp/librticonnextdds
 ADD rticonnextdds-src/librticonnextdds52-dev_5.2.0-1_amd64.deb /tmp/librticonnextdds52-dev_5.2.0-1_amd64.deb
 ADD rticonnextdds-src/rticonnextdds-tools_5.2.0-1_amd64.deb /tmp/rticonnextdds-tools_5.2.0-1_amd64.deb
 
+# Install the eProsima dependencies
+RUN apt-get update && apt-get install -y libboost-chrono-dev libboost-date-time-dev libboost-regex-dev libboost-system-dev libboost-thread-dev
+
 # Install OpenCV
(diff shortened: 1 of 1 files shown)


jacquelinekay:
I don't know much about the eProsima packages yet. Do we need to get `libboost-all-dev` or can we instead choose a subset of the Boost modules?

dirk-thomas:
A subset will likely be sufficient. Please feel free to update the PR to narrow it to specific dependencies.

jacquelinekay:
I updated the PR based on my own inspection of the eProsima repos, though I don't have their software built on my machine so I haven't tested it yet.

FastRTPS needs chrono, date_time, regex, and system, based on https://github.com/eProsima/Fast-RTPS/blob/1f744cad149a5269fb638659356b64cfb2a605c0/CMakeLists.txt#L117

timer is needed by FastBuffers: https://github.com/eProsima/Fast-Buffers/search?p=5&q=boost&utf8=%E2%9C%93

serialization is needed by the tests in DynamicFastBuffers: https://github.com/eProsima/Dynamic-Fast-Buffers/blob/0d8ebd3254655b6c6ed95d0887a2979c42a53db1/utils/performanceTests/PerformanceBoost/main.cpp#L16

I wasn't sure if we need these last two dependencies, especially the second one, but I included them just in case.

dirk-thomas:
We only use the repos specified in #116. So the `Fast-Buffers` dependencies seem to not be relevant.

dirk-thomas:
It looks like they are using way more of Boost internally: https://github.com/eProsima/Fast-RTPS/search?utf8=%E2%9C%93&q=boost&type=Code

I will mark this as in-progress until we can actually build the upstream code (see #116).

dirk-thomas:
I would like to merge this since it is a requirement for FastRTPS. Since it doesn't do any harm it shouldn't be a problem to merge it before we enable FastRTPS build by default but it makes triggering CI jobs for FastRTPS easier.

wjwwood:
It's fine by me to merge it.

PR #122, opened 2015-09-17
skipped tests should not affect the build status

dirk-thomas:
Since skipped tests are intentional (e.g. on Windows: http://ci.ros2.org/job/ros2_batch_ci_windows/419/) they should not affect the build status.


tfoote:
+1

PR #125, opened 2015-09-24
set LANG

dirk-thomas:
When building FastRTPS which contains non-ascii characters in the package manifest files it fails without this patch: http://ci.ros2.org/job/ros2_batch_ci_osx/290/

With the patch it passes that point: http://ci.ros2.org/job/ros2_batch_ci_osx/291/


wjwwood:
Lgtm

PR #132, opened 2015-10-26
make isolated builds the default option

dirk-thomas:
Being stricter in the CI jobs will prevent us from regressions like ros2/rmw_connext#105 in the future.


dirk-thomas:
And fix the generation of the launcher job...

dirk-thomas:
Please review.

tfoote:
+1

PR #133, opened 2015-10-27
adding class_loader and dependencies to repos file

wjwwood:
This adds class_loader, console_bridge, and poco_vendor to the default set of packages for CI and from source builds.

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -43,6 +43,10 @@ repositories:
     type: git
     url: https://github.com/eProsima/Fast-RTPS.git
     version: master
+  ros/class_loader:
+    type: git
+    url: https://github.com/ros/class_loader.git
+    version: ros2
   ros/geometry2:
     type: git
     url: https://github.com/ros/geometry2.git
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I updated this PR to be mergable against master adding [poco_vendor](https://github.com/ros2/poco_vendor.git) as well as [class_loader](https://github.com/ros/class_loader/tree/ros2).

I suppressed a couple of warnings in `class_loader` (https://github.com/ros/class_loader/commit/3d549eac9f103663fefa94fbf9a4158c0c7fc351) since the non-virtual destructor is intentional (see https://github.com/ros/class_loader/blob/6ccc2c0ceca6ab9f8bb60d43ed297aa99fb724cb/include/class_loader/meta_object.h#L61-L66). But before this can be merged `class_loader` needs to pass on Windows without warning and currently the package doesn't export symbols correctly.

dirk-thomas:
With ros/class_loader#41 ready this one can be considered.

dirk-thomas:
With ros/class_loader#41 ready this one can be considered for review / merge.

dirk-thomas:
Waiting for review.

PR #135, opened 2015-10-29
add cobertura to collect coverage

dirk-thomas:
Uses the data generated by ament/ament_tools#64

http://ci.ros2.org/job/ros2_batch_ci_linux/517/cobertura/


tfoote:
This is great. The metrics storage units do not make sense for the parameters but that must be legacy for the plugin. For the record the 8000000 means 80%. 

+1

dirk-thomas:
These are the default limits but they are not used anyway.

PR #140, opened 2015-11-03
define locale in packaging jobs

dirk-thomas:
This should fix http://ci.ros2.org/view/packaging/job/ros2_packaging_linux/39/


esteve:
+1

dirk-thomas:
Worked in http://ci.ros2.org/view/packaging/job/ros2_packaging_linux/40/

Issue #147, opened 2015-11-04
Select one of connext or connext dynamic on Windows Jenkins job


jacquelinekay:
Needed for rclcpp_library

wjwwood:
This job should be only Connext Static: http://ci.ros2.org/job/ros2_batch_ci_linux/557/

wjwwood:
Ok if these CI work, then I think it will be good to go:

- http://ci.ros2.org/job/ros2_batch_ci_linux/562/
- http://ci.ros2.org/job/ros2_batch_ci_osx/430/
- http://ci.ros2.org/job/ros2_batch_ci_windows/578/ (opensplice)
- http://ci.ros2.org/job/ros2_batch_ci_windows/579/ (connext static)
- http://ci.ros2.org/job/ros2_batch_ci_windows/580/ (connext dynamic)

wjwwood:
Well those all failed ... Here are more Linux and OS X:

- http://ci.ros2.org/job/ros2_batch_ci_linux/563/
- http://ci.ros2.org/job/ros2_batch_ci_osx/431/

I'm working on Windows again...

wjwwood:
Ok, this should be the final list of CI:

- http://ci.ros2.org/job/ros2_batch_ci_linux/563/
- http://ci.ros2.org/job/ros2_batch_ci_osx/437/
- http://ci.ros2.org/job/ros2_batch_ci_windows/581/ (opensplice)
- http://ci.ros2.org/job/ros2_batch_ci_windows/582/ (connext static)
- http://ci.ros2.org/job/ros2_batch_ci_windows/583/ (connext dynamic)

wjwwood:
I'll restart the CI's on https://github.com/ros2/rclcpp/pull/140 when @dirk-thomas has figured out how to keep the history.

PR #149, opened 2015-11-05
add ability to disable one of the connext implementations

wjwwood:
Connects ros2/ros2#147

### DIFF
--- a/ros2_batch_job/linux_batch/__init__.py
+++ b/ros2_batch_job/linux_batch/__init__.py
@@ -63,12 +63,6 @@ class LinuxBatchJob(BatchJob):
         def with_vendors(cmd, **kwargs):
             # Ensure shell is on since we're using &&
             kwargs['shell'] = True
-            # Extend the environment if needed.
-            env = kwargs.setdefault('env', dict(os.environ))
-            if self.args.disable_connext_static:
-                env["CONNEXT_STATIC_DISABLE"] = '1'
-            if self.args.disable_connext_dynamic:
-                env["CONNEXT_DYNAMIC_DISABLE"] = '1'
             # If the connext file is there, source it.
             if connext_env_file is not None:
                 cmd = ['.', '"%s"' % connext_env_file, '&&'] + cmd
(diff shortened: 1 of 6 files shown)


ros2_batch_job/linux_batch/__init__.py
```diff
@@ -63,6 +63,13 @@ def setup_env(self):
         def with_vendors(cmd, **kwargs):
             # Ensure shell is on since we're using &&
             kwargs['shell'] = True
+            # Extend the environment if needed.
+            env = kwargs.get('env', {})
+            if self.args.disable_connext_static:
+                env["CONNEXT_STATIC_DISABLE"] = '1'
+            if self.args.disable_connext_dynamic:
+                env["CONNEXT_DYNAMIC_DISABLE"] = '1'
+            kwargs['env'] = dict(env)
```
dirk-thomas: This block looks a bit complicated to me. At the end it set `env` anyway to be a dict. Why not do the following instead:

    env = kwargs.setdefault('env' ,{})
    if self.args.disable_connext_static:
        env["CONNEXT_STATIC_DISABLE"] = '1'
    if self.args.disable_connext_dynamic:
        env["CONNEXT_DYNAMIC_DISABLE"] = '1'

dirk-thomas: The question is does passing an empty env dict change the current behavior? Does it need to be initialized with `os.environ`?

wjwwood: I think it should be seeded with `os.environ`, 37b7893

ros2_batch_job/linux_batch/__init__.py
```diff
@@ -63,6 +63,12 @@ def setup_env(self):
         def with_vendors(cmd, **kwargs):
             # Ensure shell is on since we're using &&
             kwargs['shell'] = True
+            # Extend the environment if needed.
+            env = kwargs.setdefault('env', dict(os.environ))
+            if self.args.disable_connext_static:
+                env["CONNEXT_STATIC_DISABLE"] = '1'
```
dirk-thomas: Please use single quotes.

Same below three more times.

dirk-thomas:
lgtm

PR #155, opened 2015-11-24
fix checkout branch command when file/folder with same name exists

dirk-thomas:
Before: http://ci.ros2.org/job/ci_linux/631/
After: http://ci.ros2.org/job/ci_linux/632/


esteve:
+1

Issue #145, opened 2015-11-04
Use ccache on Jenkins


jacquelinekay:
For speed improvements.

gerkey:
#156

Issue #127, opened 2015-10-13
flaky pub / sub and request / response tests


dirk-thomas:
I ran a set of jobs as a base line:

* http://ci.ros2.org/job/ros2_batch_ci_osx/335/ **0** failing tests
* http://ci.ros2.org/job/ros2_batch_ci_windows/475/ **57** failing tests

Then I ran a set of jobs with `--ctest-args --repeat-until-fail 20`:
* http://ci.ros2.org/job/ros2_batch_ci_osx/336/ **4** failing tests
* http://ci.ros2.org/job/ros2_batch_ci_windows/476/ **59** failing tests

The arguments have no effect on Linux since the CMake version used on Ubuntu does not support this argument.

jacquelinekay:
duplicates:

ros2/rmw_connext#29 (has some logs and records of other failures)
ros2/system_tests#37 (link to RTI community post about known shared memory issue, probably not the cause since we see it in Connext and Opensplice)

wjwwood:
Should we close this since all referenced tickets are closed and we are tracking this with our nightly jobs?

dirk-thomas:
I would suggest to keep this open and aim to get it resolved in this cycle.

We happen to switch between creating tickets for failing jobs like this one and then later closing them since the problem is "covered" by the failing test. We should decide on a policy and stick to it.

dirk-thomas:
In the mean time we have updated to Xenial which has a newer CMake version and also got most of our tests stable (and for the ones which aren't retry them a few times). Closing...

PR #156, opened 2015-11-25
Add ccache to CI jobs (#145)

gerkey:
~~Not ready for review; pending CI jobs:~~  Ready for review

http://ci.ros2.org/job/ci_linux/633/
http://ci.ros2.org/job/ci_osx/519/

### DIFF
--- a/ros2_batch_job/linux_batch/__init__.py
+++ b/ros2_batch_job/linux_batch/__init__.py
@@ -27,6 +27,13 @@ class LinuxBatchJob(BatchJob):
 
     def pre(self):
         os.environ['ROS_DOMAIN_ID'] = '108'
+        # Check for ccache's directory, as installed by apt-get
+        ccache_exe_dir = '/usr/lib/ccache'
+        if os.path.isdir(ccache_exe_dir):
+            os.environ['PATH'] = ccache_exe_dir + os.pathsep +\
+                os.environ.get('PATH', '')
+        else:
+            warn('ccache does not appear to be installed; not modifying PATH')
 
     def show_env(self):
         # Show the env
--- a/ros2_batch_job/osx_batch/__init__.py
+++ b/ros2_batch_job/osx_batch/__init__.py
@@ -27,7 +27,13 @@ class OSXBatchJob(BatchJob):
 
     def pre(self):
         # Prepend the PATH with `/usr/local/bin` for global Homebrew binaries.
-        os.environ['PATH'] = "/usr/local/bin" + os.pathsep + os.environ.get('PATH', '')
+        os.environ['PATH'] = '/usr/local/bin' + os.pathsep + os.environ.get('PATH', '')
+        # Check for ccache's directory, as installed by brew
+        ccache_exe_dir = '/usr/local/opt/ccache/libexec'
+        if os.path.isdir(ccache_exe_dir):
+            os.environ['PATH'] = ccache_exe_dir + os.pathsep + os.environ.get('PATH', '')
+        else:
+            warn('ccache does not appear to be installed; not modifying PATH')
         if 'LANG' not in os.environ:
             os.environ['LANG'] = 'en_US.UTF-8'
         os.environ['ROS_DOMAIN_ID'] = '111'
--- a/job_templates/ci_job.xml.template
+++ b/job_templates/ci_job.xml.template
@@ -115,10 +115,11 @@ fi
 rm -rf ws workspace "work space"
 
 @[if os_name == 'linux']@
+mkdir -p $HOME/.ccache
 docker info
 docker build -t ros2_batch_ci linux_docker_resources
 echo "Using args: $CI_ARGS"
-docker run --privileged -e UID=`id -u` -e GID=`id -g` -e CI_ARGS="$CI_ARGS" -i -v `pwd`:/home/rosbuild/ci_scripts ros2_batch_ci
+docker run --privileged -e UID=`id -u` -e GID=`id -g` -e CI_ARGS="$CI_ARGS" -e CCACHE_DIR=/home/rosbuild/.ccache -i -v `pwd`:/home/rosbuild/ci_scripts -v $HOME/.ccache:/home/rosbuild/.ccache ros2_batch_ci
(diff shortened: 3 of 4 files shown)


gerkey:
New Linux job, after hopefully moving the PATH modification to the right place:

http://ci.ros2.org/job/ci_linux/634/

gerkey:
With this change, Linux and OSX are both using ccache, as can be seen in the console output:

http://ci.ros2.org/job/ci_linux/634/consoleFull#console-section-43
http://ci.ros2.org/job/ci_osx/519/consoleFull#console-section-43

Now, each of those jobs shows a single test failure, both related to services.   Looking over the test history for the jobs, I'm inclined to say that those are the result of the non-deterministic startup behavior for which we're still looking for a proper solution (#127).

dirk-thomas:
A repeated job run should show have a faster build time. Can you confirm this for OSX?

On Linux this will likely not be the case yet since the CCache data is local to the Docker container and therefore no brieng reused across runs.

gerkey:
Second OSX run in flight:
http://ci.ros2.org/job/ci_osx/520/

Interesting point about the ccache permanence within Docker.  Without addressing that, there's pretty much no benefit on Linux.  Looks like we can share a ccache volume across containers: http://frungy.org/docker/using-ccache-with-docker/.

There are two steps:

1. Create the shared ccache volume (e.g., `docker create -v /mnt/ccache:/ccache --name ccache debian`).  I guess that should be done just once on the slave, outside of the job.  @tfoote, how should we handle that?
1. Mount the shared ccache volume into each container.  That should just be a matter of adding the right args (e.g., `-e CCACHE_DIR=/ccache --volumes-from ccache`) to the [docker run command](https://github.com/ros2/ros2/blob/ci_scripts_add_ccache/job_templates/ci_job.xml.template#L121).

gerkey:
Elapsed time at the end of the last build step, before starting the tests (which don't benefit from ccache):

* first run: 00:19:48.613
* second run: 00:11:34.938

Also, after the second run we can see the use of ccache by checking manually on the slave:

~~~
dosa:~ osrf$ ccache -s
cache directory                     /Users/osrf/.ccache
primary config                      /Users/osrf/.ccache/ccache.conf
secondary config      (readonly)    /usr/local/Cellar/ccache/3.2.3/etc/ccache.conf
cache hit (direct)                  1388
cache hit (preprocessed)            1042
cache miss                          2218
called for link                     1236
called for preprocessing             308
compile failed                         4
bad compiler arguments                 8
autoconf compile/link                 48
no input file                         24
files in cache                      4529
cache size                         159.6 MB
max cache size                       5.0 GB
~~~

gerkey:
On OSX, ccache yields a ~40% speedup (11m vs. 19m) in the build steps and a ~20% speed overall (28m vs. 35m).

dirk-thomas:
Since we don't care about the isolation aspect I don't think we need a separate volume for the CCache data. Mounting a plain folder from the host should be enough (like here https://github.com/ros-infrastructure/ros_buildfarm/blob/f7f8143dd126b82c99ef5ce31b589406c33876fb/ros_buildfarm/templates/devel/devel_job.xml.em#L154).

gerkey:
2d4543e adds (conditional) creation of a shared .ccache on the underlying host and mounts it into the container, the appropriate environment variable setting.  To test it, I've manually modified the configuration of the ci_linux job, planning to do two runs to demonstrate proper caching.  Here's the first run:

http://ci.ros2.org/job/ci_linux/636/

Once we know that it works, I'll revert that change and we can do a proper job reconfiguration.

gerkey:
On Linux the speedup between two runs:

* http://ci.ros2.org/job/ci_linux/636/
* http://ci.ros2.org/job/ci_linux/637/

is:

* build (not test): ~65% (5m vs. 14m)
* total: ~30% (20m vs 29m)

gerkey:
I reverted the manual change to the ci_linux job configuration.

This is ready to go now; feedback welcome.

ros2_batch_job/linux_batch/__init__.py
```diff
@@ -27,6 +27,13 @@ def __init__(self, args):
 
     def pre(self):
         os.environ['ROS_DOMAIN_ID'] = '108'
+        # Check for ccache's directory, as installed by apt-get
+        ccache_exe_dir = '/usr/lib/ccache'
+        if os.path.isdir(ccache_exe_dir):
+            os.environ['PATH'] = ccache_exe_dir + os.pathsep +\
+                os.environ.get('PATH', '')
+        else:
+            warn('ccache does not appear to be installed; not modifying PATH')
```
dirk-thomas: Do we need this check for Linux if we have full control over the docker container? Even if we wouldn't install CCache the prepended path wouldn't do any harm.

ros2_batch_job/osx_batch/__init__.py
```diff
@@ -27,7 +27,15 @@ def __init__(self, args):
 
     def pre(self):
         # Prepend the PATH with `/usr/local/bin` for global Homebrew binaries.
-        os.environ['PATH'] = "/usr/local/bin" + os.pathsep + os.environ.get('PATH', '')
+        os.environ['PATH'] = '/usr/local/bin' + os.pathsep +\
+            os.environ.get('PATH', '')
```
dirk-thomas: Why wrap this line? (Same for lines 35 - 36.)

Also the trailing backslash isn't necessary.

ros2_batch_job/linux_batch/__init__.py
gerkey: The check is intended to future-proof the system so that if, for some reason we no longer have ccache install (e.g., the Dockerfile changes), it'll be fairly obvious.

ros2_batch_job/osx_batch/__init__.py
gerkey: fixed

dirk-thomas:
+1 to merge after squashed.

dirk-thomas:
I didn't think about the packaging jobs before. They would also benefit from this improvement.

Issue #146, opened 2015-11-04
Build binary packages in nightly jobs


jacquelinekay:
In addition to running all automated tests.

wjwwood:
Just to clarify are the "automated tests" separate tests for the packaging or the tests within the normal packages (the same tests we run in CI)? Currently I think the packaging job doesn't build tests, but I could be wrong about that.

jacquelinekay:
"automated tests" refers to the CI job that runs ctests. I am trying to communicate the takeaway I got from our meeting, which is that we should run packaging jobs nightly *in addition* to the current tests that are run nightly.

wjwwood:
Roger.

dirk-thomas:
Done in #157.

PR #157, opened 2015-11-25
Add nightly versions of packaging jobs, #146

gerkey:
I have no idea how to test this change.  Do we just run it, see what happens, then revert if something went wrong?

Feedback most welcome.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -131,6 +131,9 @@ def main(argv=None):
         # configure packaging job (skip non-opensplice Windows packaging jobs)
         if not os_name.startswith('windows') or 'opensplice' in os_name:
             job_name = 'packaging_' + os_name
+            # Also make it nightly, as a sanity check
+            job_data['time_trigger_spec'] = '0 12 * * *'
+            job_data['mailer_recipients'] = 'ros@osrfoundation.org'
             job_config = expand_template('packaging_job.xml.template', job_data)
             configure_job(jenkins, job_name, job_config, **jenkins_kwargs)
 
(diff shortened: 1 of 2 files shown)


tfoote:
Yeah, you could spin up a test farm to test it, but otherwise you deploy and see how it looks.

dirk-thomas:
Why does it need to be a separate set of jobs? Can we not just trigger the existing ones nightly (beside manually)?

gerkey:
@dirk-thomas , do you mean that I could add the `time_trigger_spec` setting to the existing jobs, then it would be both nightly and triggerable on demand?  I've never used Jenkins in this way, so I don't know what's possible.

dirk-thomas:
Yes, the time-based trigger is orthogonal to triggering it manually.

gerkey:
Then I would seem to have two options regarding email notification (if I'm interpreting the code correctly): send email to `ros@` for all packaging jobs, or don't send email to anyone on the nightly packaging jobs.  I'm inclined to go with no email.  Thoughts?

dirk-thomas:
I would suggest the same notification setting as for the CI jobs. They notify ros@... when they fail or are unstable (and for the first stable build after a non-stable period). Otherwise nobody will notice if these jobs are not green.

gerkey:
cf5d1e8

dirk-thomas:
+1

dirk-thomas:
I had to add the necessary snippet to the job template to make this work: cc26353b00074e3d5dbb620fab41bba86c120454

Now +1.

gerkey:
Thanks for the fix!

dirk-thomas:
This patch was incomplete and didn't enable any email notification. Follow up ticket #183.

PR #158, opened 2015-11-25
add ccache to packaging jobs

gerkey:
Not ready for review.

CI jobs:

* http://ci.ros2.org/view/packaging/job/packaging_osx/21
* http://ci.ros2.org/view/packaging/job/packaging_linux/45

We won't see any speedup on the Linux job yet (though we should see ccache being used), because it needs to be reconfigured (which we can do after this PR is merged).

### DIFF
--- a/job_templates/packaging_job.xml.template
+++ b/job_templates/packaging_job.xml.template
@@ -99,10 +99,11 @@ rm -rf ws workspace
 
 @[if os_name == 'linux']@
 export CI_ARGS="$CI_ARGS --ros1-path /opt/ros/indigo"
+mkdir -p $HOME/.ccache
 docker info
 docker build -t ros2_packaging linux_packaging_docker_resources
 echo "Using args: $CI_ARGS"
-docker run --privileged -e UID=`id -u` -e GID=`id -g` -e CI_ARGS="$CI_ARGS" -i -v `pwd`:/home/rosbuild/ci_scripts ros2_packaging
+docker run --privileged -e UID=`id -u` -e GID=`id -g` -e CI_ARGS="$CI_ARGS" -e CCACHE_DIR=/home/rosbuild/.ccache -i -v `pwd`:/home/rosbuild/ci_scripts -v $HOME/.ccache:/home/rosbuild/.ccache ros2_packaging
 @[else]@
 export CI_ARGS="$CI_ARGS --ros1-path /Users/osrf/indigo/install_isolated"
 echo "Using args: $CI_ARGS"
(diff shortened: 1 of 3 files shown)


gerkey:
Those jobs failed because the packaging job configuration is out of date with respect to the batch job script.  I think that I've fixed that in 25b4de5.  @dirk-thomas , can you have a look at that, and if you like it, then reconfigure the jobs for me, so that I can continue testing?

dirk-thomas:
I just deployed the current state of this branch to ci.ros2.ros.

gerkey:
Thanks.  The problem wasn't actually in the job configuration (though that change is still a good one, to expose those parameters for future use), but rather in the logic in the main part of the batch job, specifically:

https://github.com/ros2/ros2/blob/ci_scripts/ros2_batch_job/__main__.py#L82

where, conditioned on the value of `skip_connext`, certain fields can be added to the parser object.  Then those fields are referenced later, unconditionally.

I set `skip_connext` to `False` in the packaging job to avoid this problem.  Longer term, we might consider just removing that option, and instead relying on the more specific build parameters.

New jobs:

* http://ci.ros2.org/view/packaging/job/packaging_osx/23
* http://ci.ros2.org/view/packaging/job/packaging_linux/47

dirk-thomas:
Instead of "allowing" the package jobs to run for Connext (3e7e2e868bd602f9834e2fc1f7d4c29d0076ea23) I think the arg handling should always set the new values (`disable_connext_static`, `disable_connext_dynamic`) as it already does for the `connext` argument: https://github.com/ros2/ros2/blob/321936db046aef9ae91feb464f638f41153bf30c/ros2_batch_job/__main__.py#L116

gerkey:
I thought about that, but decided that one day, we hopefully will be able to package for Connext (and other implementations), and so we might as well have the option available (with the defaults set to good values for the current situation).

Also, note that 3e7e2e8 is necessary no matter what we "allow," unless we also change the later code to conditionally check for the presence of fields like `disable_connext_static`: https://github.com/ros2/ros2/blob/ci_scripts/ros2_batch_job/__main__.py#L171.

gerkey:
Now the ccache part is working, but the packaging job fails because the build of the ros1_bridge is broken.  I'm looking into it.

dirk-thomas:
+1 to merge (after squashed)

gerkey:
With ros2/ros1_bridge#19, the build of the bridge is fixed, and now the Linux and OSX packaging jobs are happily using cccache:

* http://ci.ros2.org/view/packaging/job/packaging_osx/25
    * Speedup: 60% (10m vs 26m)
* http://ci.ros2.org/view/packaging/job/packaging_linux/49
    * Speedup: 35% (18m vs 28m)

Note that the jobs should be reconfigured after this is merged.

gerkey:
@dirk-thomas , can you please reconfigure the jobs?  You might consider merging #157 first.

dirk-thomas:
Done.

Issue #153, opened 2015-11-12
Add support for tf in ROS2


tfoote:
- [x] ROS agnostic core library
- [x] recreate ROS wrapper/helper scripts in ROS2

tfoote:
Goal finish library by the 4th

PR #159, opened 2015-12-06
add libconsole-bridge-dev to ci environment

tfoote:
This adds libconsole-bridge into the linux docker image. 

We need to add it to the osx and windows builds too.

For OSX it's available via brew. I don't know how to resolve it for Windows besides saying to install it from source. 

Connects to #153

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -23,7 +23,7 @@ RUN curl --silent http://packages.osrfoundation.org/gazebo.key | sudo apt-key ad
 RUN apt-get update && apt-get install -y build-essential ccache cmake pkg-config python3-empy python3-setuptools python3-vcstool
 
 # Install build and test dependencies of ROS 2 packages.
-RUN apt-get update && apt-get install -y clang-format-3.4 cppcheck git pyflakes python3-coverage python3-mock python3-nose python3-pep8 uncrustify
+RUN apt-get update && apt-get install -y clang-format-3.4 cppcheck git libconsole-bridge-dev pyflakes python3-coverage python3-mock python3-nose python3-pep8 uncrustify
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Is this necessary for tf? For that it would be fine for now but for ROS 2 in general I am not sure if that is the right foundation.

tfoote:
tf needs some logging library. 

We haven't addressed logging and we should probably talk about this in the next meeting. But this has been the recommended way to do ROS independent logging that can be integrated well with ROS console logging. I think it would be a mistake not to have it be a first class citizen for ROS2.

wjwwood:
@tfoote this will not work, `console_bridge` needs changes to be used by ROS2, which I made in the `ros2` branch on the `console_bridge` repo: https://github.com/ros/console_bridge/tree/ros2

And I proposed in this pr for `class_loader`: https://github.com/ros2/ros2/pull/133

tfoote:
I can bring in console_bridge to the build tree from #133 is that what we want to do? Do we expect this to go upstream? If so we should be pushing on that now.

dirk-thomas:
Console bridge does currently not use any namespace which make it collide with other code. This was filled quite some time ago but couldn't be addressed in a non-breaking way (ros/console_bridge#18). I think we shouldn't continue to use it as-is in ROS 2 but address the problem.

wjwwood:
I think we should take the changes I made and make the ones @dirk-thomas pointed out and push for them upstream, but realistically that won't happen for a while. This isn't on my critical path though, so I haven't been pushing on #133.

tfoote:
I've pulled @wjwwood 's version from #133 into #160 and am rerunning tests.

tfoote:
console bridge is in the workspace in #160

PR #165, opened 2015-12-10
Added package for git-lfs

esteve:
This installs git-lfs and initializes it properly.

Connects to ros2/demos#43

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -11,7 +11,7 @@ ENV LANG en_US.UTF-8
 # Get curl for fetching the repo keys.
 RUN apt-get install -y curl
 
-# Get https transport for APT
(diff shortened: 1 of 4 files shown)


tfoote:
lgtm. I'm not a fan of adding the `.` to all the comments. They're technically sentences but they read more like section headers to me which wouldn't have punctuation normally.

Have we installed this on the osx and windows slaves and added it to the docs for setting up the environment for all 3?

esteve:
@tfoote I added the period at the end of the sentences that didn't have it, but I don't have an opinion whether one or the other is better, just wanted to make it consistent. Should I remove it from all the sentences?

esteve:
I've created a ticket to document Git-LFS: https://github.com/ros2/ros2/issues/173

PR #170, opened 2015-12-17
use ignore files to blacklist not used rmw related packages

dirk-thomas:
http://ci.ros2.org/job/ci_linux/755/ successfully ignores all Connext related packages.

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -16,6 +16,7 @@ import argparse
 import os
 import platform
 import sys
+from xml.etree import ElementTree
 
 # Make sure we're using Python3
 assert sys.version.startswith('3'), "This script is only meant to work with Python3"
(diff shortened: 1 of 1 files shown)


ros2_batch_job/__main__.py
```diff
@@ -154,6 +155,37 @@ def build_and_test(args, job):
...
+            with open(manifest_file, 'r') as h:
+                content = h.read()
+            try:
+                root = ElementTree.fromstring(content)
+            except Exception:
+                print("The manifest '%s' contains invalid XML" % manifest_file, file=sys.stderr)
+                raise
+            name_element = root.find('./name')
+            if name_element is None:
+                assert False, "The manifest '%s' doesn't contain a name tag" % manifest_file
```
wjwwood: Why do custom XML parsing here? Why not use the existing manifest parser?

wjwwood:
Other than the comment lgtm.

ros2_batch_job/__main__.py
dirk-thomas: Since this is before we have done anything to the workspace we do not know about that Python package at this point. We would need to hard code a relative path to `ament_package`.

wjwwood: Ok, just seems fragile. But think the alternative is probably worse. +1

tfoote:
+1, it's probably ok to stick with the custom parsing, hard coding the path inside is also not great.

PR #171, opened 2015-12-17
use ament list_packages to find to-be-blacklisted packages

dirk-thomas:
This updates #170 and uses `ament list_packages` in order to also catch packages without manifest. The dependency on the `ament_tools` path is "ok" since it is already being using to invoke the build itself.

http://ci.ros2.org/job/ci_linux/775/


wjwwood:
lgtm

PR #172, opened 2015-12-17
add fastrtps option

dirk-thomas:
Building with FastRTPS: http://ci.ros2.org/job/ci_linux/777/
Building without FastRTPS: http://ci.ros2.org/job/ci_linux/780/


esteve:
+1

gerkey:
lgtm (in the sense that it looks to me as good as the surrounding code, which in this part of our system, isn't exactly pretty :)

PR #175, opened 2015-12-18
store current branch before changing and rebasing test branch

wjwwood:
This is needed as soon as we have repositories who's default branch is not `master`, i.e. `console_bridge` or `tf2`.

Please review!


wjwwood:
Here is a Linux job that I'll cancel if it gets past this code change: http://ci.ros2.org/job/ci_linux/813/

wjwwood:
Just kidding ignore that Linux job, this OS X job will test it: http://ci.ros2.org/job/ci_osx/679/

dirk-thomas:
I just committed a mini style change.

LGTM.

PR #176, opened 2015-12-18
fix blacklisting in packaging jobs

dirk-thomas:
Seems to pass the problematic part: http://ci.ros2.org/view/packaging/job/packaging_osx/54/


wjwwood:
lgtm

PR #177, opened 2015-12-18
fix apt sources.list filename

tfoote:
This fixes the linux packaging job. 

The branch is under test here: http://ci.ros2.org/view/packaging/job/packaging_linux/81/


wjwwood:
+1 (looks like a copy-paste error from the batch ci dockerfile)

dirk-thomas:
+1

PR #183, opened 2016-01-20
fix notification for package jobs

dirk-thomas:
This fixes the notification for packaging jobs which were supposed to be added in #157.


wjwwood:
+1

PR #196, opened 2016-02-19
keep venv outside of venv

dirk-thomas:
Otherwise shebang lines use a python interpreter which has spaces in the path and spaces are not supported in shebang lines.

Before: http://ci.ros2.org/job/ci_linux/1024/
After: http://ci.ros2.org/job/ci_linux/1025/


wjwwood:
lgtm

PR #201, opened 2016-03-04
more nightly jobs

dirk-thomas:
I renamed the nightly jobs with a suffix `_repeated` (and added redirects to the Apache config).

Beside that this PR adds nightly jobs with `Debug` and `Release` (which are both not repeating the tests). With these we will have a clear history of test failures and warnings for both build types.

Timing wise the repeated jobs are triggered last.


wjwwood:
lgtm

PR #202, opened 2016-03-07
shorted nighty Windows job names

dirk-thomas:


jacquelinekay:
"nighty" :)

+1. I suspect if we ever add messages with very long names, we will run into this long path name problem again. But this is a good measure to take for now.

PR #205, opened 2016-03-18
Add ARM support

gerkey:
This PR adds support for building 32-bit and 64-bit ARM on the ROS 2 build farm.

Notes:

* I used `armhf` and `aarch64` to refer to the two platforms. That pair is one of the many competing naming schemes that are in use. I'm happy to rename if there's some other scheme that we're using.
* We're building for ARM by using docker images that run everything under qemu.
* `git` doesn't work inside `qemu`. As a result we need to pull the code outside docker and mount it in. To support this use case, I added the option `--src-mounted` to the `run_ros2_batch.py` script. In that case, it skips all the `vcs`-related steps. The config for the ARM jobs does the pulling and mounting before running the container.
* I'm only testing so far with OpenSplice because we have armhf and aarch64 debs for it. We can expand to other DDS vendors over time.
* I'm only creating manual CI jobs, which are not launched by the `ci_launcher` job. We can add other jobs over time, but because of the length of time required to run ARM/qemu builds, we'll probably never want to use them for pre-merge CI testing.
* A not-strictly-related, but useful, change in this PR is modifying the jobs to use labels (e.g., `linux`) rather than specific machine names (e.g., `linux_slave_on_master`). That way, e.g., a linux job will run on the first available linux machine. I've changed all the slaves to have the right labels.

The first armhf build is here: http://ci.ros2.org/job/ci_linux-armhf/4. So far, it's gotten through the build successfully and is running tests. I'll kick off aarch64 when armhf done.

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -121,20 +121,28 @@ def main(argv=None):
         job_config = expand_template('packaging_job.xml.template', job_data)
         configure_job(jenkins, job_name, job_config, **jenkins_kwargs)
 
+        # keeping the paths on Windows shorter
+        os_name = os_name.replace('windows', 'win')
         # configure nightly triggered job
         job_name = 'nightly_' + os_name + '_debug'
+        if os_name == 'win':
+            job_name = job_name[0:-2]
         job_data['cmake_build_type'] = 'Debug'
         job_config = expand_template('ci_job.xml.template', job_data)
         configure_job(jenkins, job_name, job_config, **jenkins_kwargs)
 
         # configure nightly triggered job
         job_name = 'nightly_' + os_name + '_release'
+        if os_name == 'win':
+            job_name = job_name[0:-4]
         job_data['cmake_build_type'] = 'Release'
         job_config = expand_template('ci_job.xml.template', job_data)
         configure_job(jenkins, job_name, job_config, **jenkins_kwargs)
 
         # configure nightly triggered job with repeated testing
         job_name = 'nightly_' + os_name + '_repeated'
+        if os_name == 'win':
+            job_name = job_name[0:-5]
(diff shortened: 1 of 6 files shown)


tfoote:
I ran armhf and aarch64 overnight and they ran to completion. A few compiler warnings and several failing tests to look into, but it did not hang as we've sometimes seen before. 

http://ci.ros2.org/job/ci_linux-aarch64/6/
http://ci.ros2.org/job/ci_linux-armhf/13/

dirk-thomas:
The patch itself looks good to me. We can always iterate on it later.

One concern I have is the significant copy-n-paste of the Docker templates. I think this should be refactored to avoid the duplication before being merged.

dirk-thomas:
Since we plan some infrastructure changes for the Alpha 7 cycle it would be important to get this merged (since it is already deployed). Otherwise it makes it much more difficult to make any further changes to the jobs.

gerkey:
After discussing with @dirk-thomas , the plan is to collapse the three Linux `Dockerfile`s into one and modify the jobs to use a combination of `sed` and docker build args to change their behavior appropriately.

gerkey:
I was able to remove both the armhf and aarch64 docker directories, so that all 3 linux builds use the same `Dockerfile` and `entrypoint.sh`. That's being tested now, but feel free to review this PR for a future merge.

dirk-thomas:
In the future it would be good if the arm jobs would also support passing a custom branch, doing merge builds with the default branch, and print the repository information. Currently neither of that happens since `vcs import` is being invoked directly for the arm jobs.

gerkey:
The test builds on x86 and armhf are progressing fine, so I'm going to merge this. I'll keep an eye on the builds this weekend and revert if anything is broken.

PR #208, opened 2016-03-25
allow concurrent CI builds

dirk-thomas:
With multiple executors per platform we need to enable concurrent builds to run more than one build at a time.

(That also allows us implicitly to run a job on the executor you want. Just trigger a job twice and abort the undesired one :wink:)

### DIFF
--- a/job_templates/ci_job.xml.template
+++ b/job_templates/ci_job.xml.template
@@ -67,7 +67,7 @@
     </hudson.triggers.TimerTrigger>
   @
 @[end if]</triggers>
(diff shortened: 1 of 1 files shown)


tfoote:
+1

dirk-thomas:
While its possible to trigger multiple builds now it seems they are still not fully concurrent. See http://ci.ros2.org/job/ci_osx/900/console#console-section-6

    Publish Cobertura Coverage Report is waiting for a checkpoint on ci_osx #899

PR #216, opened 2016-04-26
use Xenial instead of Trusty

dirk-thomas:
This should only be merged once we switched to Xenial.

### DIFF
--- a/linux_docker_resources/Dockerfile
+++ b/linux_docker_resources/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:trusty
+FROM ubuntu:xenial
 MAINTAINER William Woodall <william@osrfoundation.org>
 ARG PLATFORM=x86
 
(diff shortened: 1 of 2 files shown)


wjwwood:
There was some discussion here on the RTI forums about 16.04 support:

https://community.rti.com/forum-topic/ubuntu-1604-support

Apparently there were some issues, but it wasn't clear if they got them resolved or not. Might be worth investigating before deciding to jump our farm over to 16.04.

dirk-thomas:
I am using Xenial since mid of March and haven't seen any issue while working on ROS 2.

dirk-thomas:
The latest build (http://ci.ros2.org/job/ci_linux/1430/) looks good except that the `rclpy` tests are failing.

dirk-thomas:
The `rclpy` test failures are being resolved with ros2/rosidl#151.

jacquelinekay:
+1

dirk-thomas:
It might be a good idea to install ROS 1 Kinetic instead of failing to find Indigo... 647ace8ce6e5c1ab551787997c357f1e1266cecd

dirk-thomas:
And `rosbag` doesn't need to be patched for `lz4` anymore: 91962ee562e64b648ab6d097642d2c42e454e669

PR #222, opened 2016-05-12
install build dependency for new uncrustify version

dirk-thomas:
Connect to ament/uncrustify#13


dirk-thomas:
This will be obsolete with upstream starting to use CMake after version 0.63.

PR #229, opened 2016-06-04
change connext default to off

gerkey:
This change is already deployed on the farm.


create_jenkins_job.py
```diff
@@ -67,9 +67,9 @@ def main(argv=None):
         'ci_scripts_default_branch': args.ci_scripts_default_branch,
         'time_trigger_spec': '',
         'mailer_recipients': '',
-        'use_connext_default': 'true',
-        'disable_connext_static_default': 'false',
-        'disable_connext_dynamic_default': 'false',
+        'use_connext_default': 'false',
+        'disable_connext_static_default': 'true',
```
dirk-thomas: Line 60 is sufficient. This line and the following one don't need to change and make it more effort to enable Connext on demand.

wjwwood: I agree. As-is you'd have to check three boxes to get the old behavior. Setting `use_connext_default` to `false` should be sufficient.

gerkey: d4094ca2db92db923c36915d08c49d3f72acc731

Also pushed to the farm.

wjwwood:
Thanks @gereky, lgtm!

PR #231, opened 2016-06-06
Make the launcher job also inherit the defaults used for other jobs

gerkey:
@dirk-thomas discovered that, following #229, the launcher job was still using the old defaults. That's because that job is configured manually in the job creation script in a manner that doesn't use those defaults, and then the launcher job template has its own hard-coded defaults. This PR makes the launcher job inherit the same defaults used elsewhere.

It's deployed on the farm, which resulted, as expected, in this single-line change to the launcher job config (changing connext from default on to default off):

```
Updating job 'ci_launcher'
    <<<
    --- remote config
    +++ new config
    @@ -38 +38 @@
    -          <defaultValue>true</defaultValue>
    +          <defaultValue>false</defaultValue>
    >>>
```

### DIFF
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -170,12 +170,11 @@ def main(argv=None):
         os_specific_data[os_name] = dict(data)
         os_specific_data[os_name].update(os_configs[os_name])
         os_specific_data[os_name]['job_name'] = 'ci_' + os_name
-    job_data = {
-        'ci_scripts_default_branch': args.ci_scripts_default_branch,
-        'label_expression': 'master',
-        'os_specific_data': os_specific_data,
-        'cmake_build_type': 'None',
-    }
+    job_data = dict(data)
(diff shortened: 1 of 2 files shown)


create_jenkins_job.py
```diff
@@ -170,12 +170,11 @@ def main(argv=None):
...
-        'ci_scripts_default_branch': args.ci_scripts_default_branch,
-        'label_expression': 'master',
-        'os_specific_data': os_specific_data,
-        'cmake_build_type': 'None',
-    }
+    job_data = dict(data)
+    job_data['ci_scripts_default_branch'] = args.ci_scripts_default_branch
+    job_data['label_expression'] = 'master'
+    job_data['os_specific_data'] = os_specific_data
+    job_data['cmake_build_type'] = 'None'
```
dirk-thomas: Why change this block?

dirk-thomas: Nvm. Had to read twice.

dirk-thomas:
+1

Issue #230, opened 2016-06-04
Add coverage builds


gerkey:
Start with gcov/lcov, reporting through a Jenkins plugin (e.g., Clover).

dirk-thomas:
We might want to try the `Cobertura` plugin first since we already use that for Python coverage. And use `gcovr` since that generates coberture compatible result files.

wjwwood:
Should this be closed or remain open until the follow up issues are resolved?

gerkey:
The basics are there; we'll ticket problems and enhancements separately.

PR #234, opened 2016-06-15
Enable C++ coverage reports using gcov

jacquelinekay:
In progress.

Right now coverage is toggled with a new parameter on Jenkins, CI_COVERAGE, which is off by default. I can change that to a different path (e.g. a different job template type) if we think that's appropriate.

This build should show what the coverage results will look like on Jenkins.

http://ci.ros2.org/job/ci_linux/1457/

Connects to #230

### DIFF
--- a/ros2_batch_job/__main__.py
+++ b/ros2_batch_job/__main__.py
@@ -55,6 +55,7 @@ pip_dependencies = [
     'vcstool',
 ]
 
+gcov_flags = " -fprofile-arcs -ftest-coverage "
 
 def main(sysargv=None):
     args = get_args(sysargv=sysargv)
@@ -116,6 +117,9 @@ def get_args(sysargv=None, skip_white_space_in=False, skip_connext=False, add_ro
     parser.add_argument(
         '--src-mounted', default=False, action='store_true',
         help="src directory is already mounted into the workspace")
+    parser.add_argument(
+        '--coverage', default=False, action='store_true',
+        help="enable collection of coverage statistics")
 
     argv = sysargv[1:] if sysargv is not None else sys.argv[1:]
     argv, ament_args = extract_argument_group(argv, '--ament-args')
@@ -132,8 +136,52 @@ def get_args(sysargv=None, skip_white_space_in=False, skip_connext=False, add_ro
     return args
 
 
+def process_coverage(args, job):
+    print('# BEGIN SUBSECTION: coverage analysis')
+    # Collect all .gcda files in args.workspace
+    ament_py = get_ament_script(args.sourcespace)
+    output = subprocess.check_output(
+        [job.python, '-u', ament_py, 'list_packages', args.sourcespace])
+    for line in output.decode().splitlines():
+        package_name, package_path = line.split(' ', 1)
+        package_build_path = os.path.join(args.buildspace, package_name)
+        gcda_files = []
+        for root, dirs, files in os.walk(package_build_path):
+            gcda_files.extend(
+                    [os.path.abspath(os.path.join(root, f))
+                        for f in files if f.endswith('.gcda')])
+        if len(gcda_files) is 0:
+            continue
+        print(gcda_files)
+
+        # Run one gcov command for all gcda files for this package.
+        subprocess.run(['gcov', '--preserve-paths'] + gcda_files, check=True)
+        # Remove coverage files that did not originate in the workspace
+        for cov_file in (f for f in os.listdir(package_build_path)
+                         if f.endswith('.gcov')):
+            cov_path = cov_file.replace('#', '/')
+            # TODO Better path filter here
+            if not cov_path.beginswith(os.path.abspath(args.workspace)):
+                os.remove(os.path.join(root, cov_file))
+        # Write one report for the entire package.
+        # cobertura plugin looks for files of the regex *coverage.xml
+        outfile = os.path.join(package_build_path, package_name + '.coverage.xml')
+        print('Writing coverage.xml report at path {}'.format(outfile))
+        # -e /usr  Ignore files from /usr
+        # -xml  Output cobertura xml
+        # -output=<outfile>  Pass name of output file
+        # -g  use existing .gcov files in the directory
+        subprocess.run(
+            ['gcovr', '-e', '/usr', '--xml', '--output=' + outfile, '-g'],
+            check=True)
+
+    print('# END SUBSECTION')
+    return 0
+
+
 def build_and_test(args, job):
     ament_py = get_ament_script(args.sourcespace)
+    coverage = args.coverage and args.os == 'linux'
 
     print('# BEGIN SUBSECTION: ament build')
     # Now run ament build
@@ -144,9 +192,14 @@ def build_and_test(args, job):
         '"%s"' % args.sourcespace
     ] + (['--isolated'] if args.isolated else []) +
         (
-            ['--cmake-args', '-DCMAKE_BUILD_TYPE=' + args.cmake_build_type]
+            ['--cmake-args', '-DCMAKE_BUILD_TYPE=' +
+                args.cmake_build_type + ' -- ']
             if args.cmake_build_type else []
-        ) + args.ament_args, shell=True)
+        ) + args.ament_args +
+        (['--ament-cmake-args -DCMAKE_CXX_FLAGS="${CMAKE_CXX_FLAGS} ' +
+            gcov_flags + ' " -DCMAKE_C_FLAGS="${CMAKE_C_FLAGS} ' +
+            gcov_flags + '" -- ']
+            if coverage else []), shell=True)
     print('# END SUBSECTION')
 
     print('# BEGIN SUBSECTION: ament test')
@@ -171,6 +224,9 @@ def build_and_test(args, job):
     )
     info("ament.py test_results returned: '{0}'".format(ret_test_results))
     print('# END SUBSECTION')
+    if coverage:
+        process_coverage(args, job)
+
     # Uncomment this line to failing tests a failrue of this command.
     # return 0 if ret_test == 0 and ret_testr == 0 else 1
     return 0
--- a/create_jenkins_job.py
+++ b/create_jenkins_job.py
@@ -73,6 +73,7 @@ def main(argv=None):
         'use_fastrtps_default': 'true',
         'use_opensplice_default': 'false',
         'ament_args_default': '',
+        'enable_c_coverage_default': 'false',
     }
 
     jenkins = connect(args.jenkins_url)
@@ -143,6 +144,14 @@ def main(argv=None):
         job_data['cmake_build_type'] = 'Debug'
         job_config = expand_template('ci_job.xml.template', job_data)
         configure_job(jenkins, job_name, job_config, **jenkins_kwargs)
+
+        # configure nightly coverage job on Linux only
+        if os_name == 'linux':
+            job_name = 'nightly_' + os_name + '_coverage'
+            job_data['cmake_build_type'] = 'Debug'
+            job_data['enable_c_coverage_default'] = 'true'
+            job_config = expand_template('ci_job.xml.template', job_data)
+            configure_job(jenkins, job_name, job_config, **jenkins_kwargs)
 
         # configure nightly triggered job
         job_name = 'nightly_' + os_name + '_release'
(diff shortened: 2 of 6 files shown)


jacquelinekay:
Whoops, I didn't change the option on Jenkins for making Cobertura look at all of the `coverage.xml` output files. Here we go again:

ci.ros2.org/job/ci_linux/1458/

gerkey:
Looks like something went wrong during the `gcov` invocations, perhaps before more `.xml` content was generated for Jenkins to report on:
```
22:34:47 File '/usr/include/c++/5/exception'
22:34:47 Lines executed:0.00% of 1
22:34:47 Creating '#usr#include#c++#5#exception.gcov'
22:34:47 
22:34:47 Exception ignored when trying to write to the signal wakeup fd:
22:34:47 BlockingIOError: [Errno 11] Resource temporarily unavailable
22:34:48 Exception ignored when trying to write to the signal wakeup fd:
22:34:48 BlockingIOError: [Errno 11] Resource temporarily unavailable
22:34:48 # END SUBSECTION
```

jacquelinekay:
indeed, I'm trying to reproduce that on my local Docker instance now.

jacquelinekay:
From my local testing, the BlockingIOErrors are concerning, but the coverage.xml files are being created under the CMakeFiles directory. So I suspect there is something wrong with how the Cobertura Jenkins plugin is configured to process those files, since the coverage report doesn't have any information about C++ code:

http://ci.ros2.org/job/ci_linux/1458/cobertura/

gerkey:
Can you check in the Jenkins workspace for the coverage files? I think that you'd need to grab it immediately after you build, to stop it from being overwritten by the next job.

jacquelinekay:
I will do that next time I run a job for this branch, since a number of jobs have happened since the last run.

ros2_batch_job/__main__.py
```diff
@@ -131,9 +135,40 @@ def get_args(sysargv=None, skip_white_space_in=False, skip_connext=False, add_ro
...
+                # Remove coverage files
+                for cov_file in (f for f in os.listdir(root) if f.endswith('.gcov')):
+                    cov_path = cov_file.replace('#', '/')
+                    if not cov_path.beginswith(args.workspace):
+                        os.remove(os.path.join(root, cov_file))
+                # We expect this to generate .gcov files
+                # Save coverage output to filename.gcda.dir
+                # or figure out a better regex to pass to the cobertura plugin
+                outfile = os.path.join(root, gcda_file + '.coverage.xml')
+                subprocess.call(['gcovr', '--xml', '--output=' + outfile, '-g'])
```
dirk-thomas: You need to check the return code of the subprocess.

You might also want to print something to know from the log which file this was supposed to be generating.

jacquelinekay: Yeah, in my local testing I am using `check=True` to the subprocesses but that does not propagate the exception up to this context. So I don't think the subprocess calls are the direct cause of the BlockingIOErrors seen in the output.

dirk-thomas:
Since the generated `coverage.xml` files are not in a direct subfolder of the package specific build folder the regex to match these files won't find them. Instead of:

    ws/build*/*/coverage.xml

you need something like this:

    ws/build*/**/coverage.xml

jacquelinekay:
Thanks for the tip. I figured it was something wrong with the Jenkins plugin regex. My original thought was that * in the front of a filename, e.g. `*coverage.xml`, would not get resolved properly, so I changed the name of all coverage reports to "coverage.xml" and saved them to uniquely named directories, e.g. `<executable name>.dir/coverage.xml`. I am going to revert this scheme with the different directories because it's quite inefficient and call the reports `<executable name>.coverage.xml`. Will the regex `ws/build*/**/*coverage.xml` work in this case?

jacquelinekay:
Alright, here's the first successful coverage report on Jenkins:

http://ci.ros2.org/job/ci_linux/1471/cobertura/

dirk-thomas:
That looks like a lot of numbers and a lot of green :wink:

It currently shows coverage for stuff in `/usr/include/...`. Can they be removed?

The top level names start with `.home.rosbuild.ci_scripts.ws.src`. Can this prefix be stripped to only start on the source space? Even better would be to just use the package name as the prefix.

Also I can't find some of our packages, e.g. the `intra_process_demo`.

ros2_batch_job/__main__.py
```diff
@@ -132,8 +136,45 @@ def get_args(sysargv=None, skip_white_space_in=False, skip_connext=False, add_ro
...
+        cmake_artifacts_path = os.path.join(package_build_path, "CMakeFiles")
+        if not os.path.exists(cmake_artifacts_path):
+            print(
+                'Could not find CMakeFiles for package {}, skipping.'.format(
+                    package_name))
+            continue
+        for root, dirs, files in os.walk(cmake_artifacts_path):
+            for gcda_file in (f for f in files if f.endswith('.gcda')):
+                subprocess.run(
+                    ['gcov', '--preserve-paths', os.path.join(
```
dirk-thomas: Should we pass `--relative-only` here?

jacquelinekay: In my local testing with gcov, `--relative-only` produced files with no output. I think this is because the gcov artifacts (.gcda and .gcno files) end up in a different directory from the sources, so it thinks even the original sources are "relative". I haven't tried it recently on the farm though.

jacquelinekay:
I started a different job, "coverage_test", because I was messing around with xml options. After some more research I think simply passing `-e /usr` to gcovr will do the trick for ignoring standard library files.

http://ci.ros2.org/job/coverage_test/2/

gerkey:
I noticed that the C++ results don't include any conditionals coverage. Does `gcov` not support that analysis, or would passing `-b` or `-c`?

jacquelinekay:
I'll try passing `-b` to `gcovr`.

jacquelinekay:
Either we have really bad conditional coverage, or I'm doing something wrong.

Passing `-b` to gcov: http://ci.ros2.org/job/coverage_test/4/cobertura/
  Shows zero conditional coverage for all C++ packages except for the first one.
Passing `-b` to gcov and gcovr: http://ci.ros2.org/job/coverage_test/5/cobertura/
  Same as above.
Passing `-b` to just gcovr: http://ci.ros2.org/job/coverage_test/6/
  N/A reported for conditionals.

I will investigate.

gerkey:
Recognizing that I suggested investigating conditional coverage: I wouldn't get hung up on that now if it's proving to be problematic. Having the file, class, and line coverage is so much better than not having anything that it's very worthwhile to deploy as is.

How are the build times and test results when coverage is enabled?

jacquelinekay:
The last coverage build took 32 minutes on "linux 2", while the typical build on the "linux 2" slave takes around 20-24 minutes (see e.g. http://ci.ros2.org/job/ci_linux/1472/, http://ci.ros2.org/view/nightly/job/nightly_linux_debug/112/).

jacquelinekay:
I *just* realized that the coverage build tests don't build with Debug build type. Going to rerun everything to see if that changes timing/branch coverage results.

gerkey:
So it's ~+45% in build time. That's manageable enough for doing a nightly coverage build, with the option to enable coverage in a manual build when desired.

Suggestion: disable test result reporting in the standard (e.g., nightly) coverage build. Enabling coverage can cause spurious test failures, and we have other builds (many, many, other builds) to run and report on tests. The goal of the coverage builds is to report on how well the tests cover the code, not whether they're passing.

jacquelinekay:
OK, I've rearranged things so that there should be one report per package.

I also add the right gcc flags to "CMAKE_C_FLAGS" to collect coverage information for C code (e.g. rcl, C typesupport).

I'm still not sure what to do about the long path names short of stripping them directing from the files.

http://ci.ros2.org/job/coverage_test/9/

If this looks better I'll look into setting up a nightly coverage job like the other nightlies.

jacquelinekay:
I put the coverage reports in the wrong folder in the previous build so the packages didn't show up in the Jenkins web UI, relaunched here: http://ci.ros2.org/job/coverage_test/10/

gerkey:
+1 to merge and deploy. We can make further tweaks (e.g., shortening names) over time.

dirk-thomas:
Were you able to identify if the results are actually "complete" / where the results for some packages are?

tfoote:
That looks good. I suspect that a regex will likely be the best way to shorten the paths for the reports. 

The one other improvement that would be great would be to have coburtura know how to find the source so that we can get the annotated source files. But that's an enhancement we can add later. 


For example clicking into a specific file's report, like: http://ci.ros2.org/job/coverage_test/10/cobertura/ament_pep8/main_py/ 
```
Source code is unavailable. Some possible reasons are:

This is not the most recent build (to save on disk space, this plugin only keeps the most recent build’s source code).
Cobertura found the source code but did not provide enough information to locate the source code.
Cobertura could not find the source code, so this plugin has no hope of finding it.
You do not have sufficient permissions to view this file.
```

dirk-thomas:
Also the parameter string on a job seems to not wrap before the coverage flag:

```
branch: , <br/>
use_connext: true, <br/>
disable_connext_static: false, <br/>
disable_connext_dynamic: false, <br/>
use_fastrtps: true, <br/>
use_opensplice: true, <br/>
ci_branch: ci_scripts, <br/>
repos_url: , <br/>
use_whitespace: false, <br/>
isolated: true, <br/>
cmake_build_type: None, <br/>
ament_args: coverage: false
            ^
```

wjwwood:
It would be great if this was merged and deployed and other comments are addressed in follow up pr's. Can we do that? Is the latest version deployed?

gerkey:
+1 (again) to merge a working version now, and iterate on it over time.

jacquelinekay:
Yeah, I've spent a ton of time fiddling with little details like getting one report per package, etc.

I set up a testing job here: http://ci.ros2.org/job/coverage_test/ But I was going to go with Brian's suggestion of configuring a nightly job.

Since I'm going to be out of the office in the afternoon I can finish up the buildfarm config stuff tomorrow morning or someone else can take care of it. I'm less familiar with Jenkins configuration stuff but I can figure it out.

wjwwood:
So... should I merge and deploy this now? @jacquelinekay it wasn't clear to me from your last comment if that's ok or if you need to do more stuff to it before it can be merged.

jacquelinekay:
As a sanity check I'm also going to launch jobs that use this branch of ci_scripts on non-Linux builds.

jacquelinekay:
Just to make sure that this branch doesn't break other platforms:

http://ci.ros2.org/job/ci_osx/1192/
http://ci.ros2.org/job/ci_windows/1523/

After these come out "clean" this can be merged.

dirk-thomas:
It still needs to be investigated why only part of the packages show up and some are missing completely.

wjwwood:
@jacquelinekay I'm going to cancel this job since it is hung on the FastRTPS test:

http://ci.ros2.org/job/ci_osx/1192/

gerkey:
In a well-intentioned attempt to help, I just git-vomited all over this branch. Hold while I get it back into a reasonable state.

gerkey:
OK, I think that I've resurrected this branch, including a trivial formatting fix.

gerkey:
I added Jenkins configuration and deployed it. I'm running a test build now:

http://ci.ros2.org/view/nightly/job/nightly_linux_coverage/1/

If that works, I'll merge this.

gerkey:
Whoops, got my order of operations wrong there. Now that I've deployed it, I need to merge this PR because all the jobs are now having the `--coverage` argument passed to the batch build script, which isn't expecting it.

wjwwood:
You could have used the `CI_SCRIPTS_BRANCH` to test with this branch 😄.

gerkey:
Yeah, but I was trying to test the deployed result, and after deploying it, all the jobs take the coverage param and pass it through. I need a way to fork Jenkins...

gerkey:
I'm running a new coverage build and I also did a manual launch on all three platforms as a sanity check.

gerkey:
Minor fix in #243.

ros2_batch_job/__main__.py
```diff
@@ -132,8 +136,52 @@ def get_args(sysargv=None, skip_white_space_in=False, skip_connext=False, add_ro
...
+        print(gcda_files)
+
+        # Run one gcov command for all gcda files for this package.
+        subprocess.run(['gcov', '--preserve-paths'] + gcda_files, check=True)
+        # Remove coverage files that did not originate in the workspace
+        for cov_file in (f for f in os.listdir(package_build_path)
+                         if f.endswith('.gcov')):
+            cov_path = cov_file.replace('#', '/')
+            # TODO Better path filter here
+            if not cov_path.beginswith(os.path.abspath(args.workspace)):
```
dirk-thomas: `beginswith` is not a valid function. See ros2/ci#1 for the fix.

PR #236, opened 2016-06-22
add valgrind to Linux Dockerfile

dirk-thomas:
We can either install `valgrind` on the OS X slaves manually or hope that eProsima/Fast-RTPS#49 gets addressed and live with the failing tests until then.

http://ci.ros2.org/job/ci_linux/1510/


dirk-thomas:
@wjwwood has installed `valgrind` on the OS X slaves :clap:

wjwwood:
Can this be merged? lgtm

PR #243, opened 2016-06-24
fix condition on coverage variable

gerkey:
The old code was checking for non-zero-length instead of comparing against true. The result was that `--coverage` was being passed every time, which had the effect of making all Linux builds do coverage.

This change has been deployed to the farm.

### DIFF
--- a/job_templates/ci_job.xml.template
+++ b/job_templates/ci_job.xml.template
@@ -132,7 +132,7 @@ fi
 if [ -n "${CI_AMENT_ARGS+x}" ]; then
   export CI_ARGS="$CI_ARGS --ament-args $CI_AMENT_ARGS"
 fi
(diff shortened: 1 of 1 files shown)


gerkey:
Builds using the deployed code to demonstrate correct behavior:

* nightly coverage job: http://ci.ros2.org/job/nightly_linux_coverage/3/
* non-coverage Linux job: http://ci.ros2.org/job/ci_linux/1526/
* non-coverage OSX job: http://ci.ros2.org/job/ci_osx/1197/

tfoote:
+1

wjwwood:
+1

gerkey:
The builds have gotten past the point of determining whether to pass the `--coverage` arg to the batch script, and have done so correctly.

PR #245, opened 2016-06-25
Use connext binaries off RTI website

dhood:
I am yet to add an option to job parameters for choosing between debian packages and binaries off RTI website


dhood:
OK, this PR now makes the default setting on the build farm use connext off the website for linux. there is an option for going back to using the debs.

points for consideration:

1. this will work as long as the specified (hard-coded) installer bundle is hosted in the same place on the RTI website. if it moves, the jobs will fail. we will have to be responsible for updating the download link each time new releases are available (it can be automated, but this is the case for the other platforms anyway)
2. this only affects linux at the moment. from what I understood the osx and windows slaves are using connext binaries from the RTI portal with OSRF login. I propose installing the packages from http://www.rti.com/downloads/connext-files.html instead, to match the typical user experience. I can add an option for switching between the two install paths if appropriate.

linux_docker_resources/Dockerfile
```diff
@@ -76,6 +86,9 @@ RUN sudo -H -u rosbuild -- git config --global user.name "Jenkins ROS 2"
 RUN echo 'rosbuild ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
 WORKDIR /home/rosbuild
 
+# Ensure the default shell command language is bash
+RUN rm /bin/sh && ln -s /bin/bash /bin/sh
```
dirk-thomas: Why is this necessary?

job_templates/snippet/property_parameter-definition.xml.em
```diff
@@ -43,6 +43,11 @@ This tests the robustness to whitespace being within the different paths.</descr
           <defaultValue>@(disable_connext_dynamic_default)</defaultValue>
         </hudson.model.BooleanParameterDefinition>
         <hudson.model.BooleanParameterDefinition>
+          <name>CI_USE_OSRF_CONNEXT_DEBS</name>
+          <description>By setting this to True, the build will use the deb packages built by OSRF for Connext, instead of the binaries off the RTI website (applies to linux only).</description>
```
dirk-thomas: Should the option only be made available on Linux jobs? Or should we add support for OS X and Windows directly?

linux_docker_resources/Dockerfile
dhood: otherwise it will use dash and you [get 'bad substitution' errors](http://ci.ros2.org/job/ci_linux/1548/console#console-section-6) 
```
# BEGIN SUBSECTION: ament build
22:38:23 (RTI)
22:38:23 (venv)
22:38:23 ==> . "/home/rosbuild/rti_connext_dds-5.2.3/resource/scripts/rtisetenv_x64Linux3gcc4.8.2.bash" && "/home/rosbuild/ci_scripts/venv/bin/python" -u "src/ament/ament_tools/scripts/ament.py" build --build-tests --build-space "build" --install-space "install" "src" --isolated
22:38:23 /bin/sh: 4: /home/rosbuild/rti_connext_dds-5.2.3/resource/scripts/rtisetenv_x64Linux3gcc4.8.2.bash: Bad substitution
22:38:23 <== '. ../venv/bin/activate && . "/home/rosbuild/rti_connext_dds-5.2.3/resource/scripts/rtisetenv_x64Linux3gcc4.8.2.bash" && "/home/rosbuild/ci_scripts/venv/bin/python" -u "src/ament/ament_tools/scripts/ament.py" build --build-tests --build-space "build" --install-space "install" "src" --isolated' exited with return code '2'
```

dirk-thomas: Would it be sufficient to change the `ENTRYPOINT` or `CMD` rather then fiddling with these global binaries?

job_templates/snippet/property_parameter-definition.xml.em
dhood: I have the same question. it would depend on how different (if at all) the binaries off the RTI website are for those platforms.. there might not be a motivation to use anything but the website ones for OS X and windows

linux_docker_resources/Dockerfile
dhood: since bash is the default login shell for users, maybe it's not such a bad thing to change it globally at this point in the docker image? otherwise I can look into making it take effect as locally as possible

dirk-thomas: Changing it globally has the chance of side effects or everything which uses `/bin/sh`. Therefore I suggest to keep the change as local as possible.

wjwwood: Yeah, maybe just changing this to use `/bin/bash` will be enough?

https://github.com/ros2/ros2/blob/ci_scripts/linux_docker_resources/entry_point.sh#L49

I'm not sure that will be sufficient though. I don't know how the subprocess module in Python selects which shell it will use when you pass it `shell=True`, which is what I assume is happening.

Actually, I just looked it up:

From https://docs.python.org/2/library/subprocess.html#popen-constructor:
```
On Unix with shell=True, the shell defaults to /bin/sh.
```

I didn't look to see if there is another option. We could also use `shell=False` and invoke `bash` directly with `-c`?

wjwwood: According to this site: http://www.saltycrane.com/blog/2011/04/how-use-bash-shell-python-subprocess-instead-binsh/

You should be able to use the `executable` keyword argument, but that didn't seem to work for them, so they ended up using `['/bin/bash', '-c', cmd]` (which was what I was think we could do too).

dhood: @dirk-thomas I agree that it would have side effects, but I meant that we would be better representing what the typical user would have on their system. i.e. I'm arguing for the use of bash by default, and accommodating whatever side effects come up (I haven't noticed any so far)

wjwwood: I think @dirk-thomas is right that this will have side effects beyond what a normal user will experience. A normal user will have bash or zsh as a login shell yes, but according to the Python documentation, all subprocess calls with shell=True will default to `/bin/sh`. On a normal user's computer that will still be sh (or dash in the case of Linux), but with this change the Python invocations will now be using bash rather than dash. So I don't think this is the same as what users will experience.

I'd really recommend invoking bash directly in any subprocess calls that need to source the rti binary. If that is possible (I'm not sure it is).

dhood: ok, I see your point - thanks for the explanation. I'll look into it

wjwwood: I have yet another suggestion. It seems that the issue is that their shell script uses `"${BASH_SOURCE[0]}"` to get the directory of the script (which is something that we know from experience is hard to do in plain `sh`). You could replace that with the full path to the script with `sed` or something after installing the rti package. I just tried that in my Linux VM and after replacing that with the script path, I can source it in plain sh.

dhood: whereas before you got the `Bad substitution` error?

wjwwood: Yes.

wjwwood:
```
% sh
$ . rti_connext_dds-5.2.3/resource/scripts/rtisetenv_x64Linux3gcc4.8.2.bash
sh: 4: rti_connext_dds-5.2.3/resource/scripts/rtisetenv_x64Linux3gcc4.8.2.bash: Bad substitution
$ vim rti_connext_dds-5.2.3/resource/scripts/rtisetenv_x64Linux3gcc4.8.2.bash
>>>  This is where I replaced the BASH_SOURCE stuff with the full path to the script.
$ . rti_connext_dds-5.2.3/resource/scripts/rtisetenv_x64Linux3gcc4.8.2.bash
###############################################################################

         (c) Copyright, Real-Time Innovations, All rights reserved.            
                                                                               
                           RTI Connext DDS 5.2.3                       

###############################################################################
$ 
```

dhood: ah ok that's whose script we're talking about. I'm trying to decide which approach would be best. switching to bash seems more robust to connext changes but at this point changing the shell script seems like a more confined workaround

wjwwood: Yeah, I have a slight preference for modifying the script after installation. @dirk-thomas?

dirk-thomas: I also think modifying the script after installation is the best way to go.

dhood: ok thanks, will ping you when ready for review again

linux_docker_resources/entry_point.sh
```diff
@@ -42,8 +42,13 @@ case "${CI_ARGS}" in
 	dpkg -i /tmp/rticonnextdds-tools_5.2.3-1_amd64.deb
 	;;
       *)
-	echo "Install Connext binaries off RTI website..."
+	echo "Installing Connext binaries off RTI website..."
 	python3 -u /tmp/rti_web_binaries_install_script.py /tmp/rti-installer.run /home/rosbuild
+	if [ $? -ne 0 ]
+	then
+	  echo "Connext not installed correctly." >&2
+	  exit 1
```
dhood: is there a simple way to apply this sort of check to each step in the script so that jobs stop on errors?
we have an error in this script at line 25 that seems to be going unnoticed http://ci.ros2.org/job/ci_linux/1603/consoleFull#console-section-4 (job unrelated to this PR)

linux_docker_resources/Dockerfile
dhood: have done as suggested and modified the connext env script. did it in python instead of sed so that the location of the script is only used in one place.

dhood:
new job is building fine: http://ci.ros2.org/job/ci_linux/1615

wjwwood:
lgtm

dhood:
Update:

Linux with default job options: [Build Status](http://ci.ros2.org/job/ci_linux/1735/)
OS X with default job option: [Build Status](http://ci.ros2.org/job/ci_osx/1295/)

Linux with connext dynamic included: [Build Status](http://ci.ros2.org/job/ci_linux/1737/)
Windows with default job options: [Build Status](http://ci.ros2.org/job/ci_windows/1668/)

I'm going to merge this and push it to the farm (no one seems to be testing anything)

wjwwood:
+1

dirk-thomas:
I am not sure if you still need the branch. If not you can delete it using the button.

dirk-thomas:
@dhood When you deployed the changed to Jenkins did you also change the `aarch64` and `armhf` jobs? If I remember correctly they had local modification which are not covered by the checked in code.

dhood:
I didn't do anything special for those jobs: I ran the `create_jenkins_job` script with the code on this branch. So I think the local modifications you're talking about would need to be applied again

dirk-thomas:
@gerkey FYI. I think you created and modified those jobs originally. Since the configuration was not covered in this repo the modification on the jobs have been reverted. You might either want to reapply the changes from some local backup or try to look into the job config history before using the jobs again.

PR #248, opened 2016-06-27
add CMake warning parser

dirk-thomas:
Use the new regular expression / Groovy script (configured globally) to extract CMake warnings and errors.

### DIFF
--- a/job_templates/snippet/publisher_warnings.xml.em
+++ b/job_templates/snippet/publisher_warnings.xml.em
@@ -35,6 +35,9 @@
       <parserConfigurations/>
       <consoleParsers>
         <hudson.plugins.warnings.ConsoleParser>
+          <parserName>CMake</parserName>
+        </hudson.plugins.warnings.ConsoleParser>
+        <hudson.plugins.warnings.ConsoleParser>
 @[if os_name in ['linux', 'linux-armhf', 'linux-aarch64']]@
           <parserName>GNU C Compiler 4 (gcc)</parserName>
 @[elif os_name == 'osx']@


dirk-thomas:
For the record the regex looks like this:

    ^CMake (Deprecation Warning|Warning|Warning \(dev\)|Error) at (.+):(\d+) \((.+)\):\n(  .+\n)?(.+\n(  .+\n)*)?

and the Groovy script transforming the matches into warnings:

```
import hudson.plugins.analysis.util.model.Priority
import hudson.plugins.warnings.parser.Warning

String messageType = matcher.group(1)
String fileName = matcher.group(2)
String lineNumber = matcher.group(3)
String scope = matcher.group(4)
String cmakeMessage = matcher.group(5)
String callStack = matcher.group(6)

String message = ""
if (cmakeMessage) {
    message += cmakeMessage + "<br />"
}
if (callStack) {
    message += "<pre>" + callStack + "</pre>"
}

Priority priority = Priority.NORMAL
if (messageType == "Deprecation Warning") {
    priority = Priority.LOW
} else if (messageType == "Error") {
    priority = Priority.HIGH
}
return new Warning(fileName, Integer.parseInt(lineNumber), "CMake", scope, message, priority);
```

dirk-thomas:
Waiting for review...

wjwwood:
If that's a global configuration, and is not covered by the deploy script, then it should be mentioned in the setup instructions: https://github.com/ros2/ros2/blob/ci_scripts/README.md#adjusting-batch-ci-jobs (or a similar section)

Otherwise, lgtm.

dirk-thomas:
We currently have only very few things documented regarding the Jenkins master configuration. I don't think it is worth to add this particular detail.

wjwwood:
In order to set up the farm again someone would need to remember this needs to be done and find this issue and the regex/groovy script. I think it's a reasonable thing to mention in the setup README, so I added a note that at least points to your comment: https://github.com/ros2/ros2/commit/eacb1b7953ad0933c47e33cd8eeb210c90779a2a

Your comment might get out-of-date, but at least it's a starting point for someone else.

PR #250, opened 2016-06-28
fix unicode problem on Windows

dirk-thomas:
This should fix the traceback in the Windows builds: http://ci.ros2.org/view/nightly/job/nightly_win_deb/165/consoleFull#console-section-2


wjwwood:
+1, Was that an invalid character or just Windows not supporting unicode?

dirk-thomas:
It was `chr(225)` which is an `a` with "something" on top. I assume one of the latest log entries on an eProsima repo maybe?

dirk-thomas:
This didn't fix the problem: http://ci.ros2.org/job/ci_windows/1552/consoleFull#console-section-3

PR #256, opened 2016-07-11
add option to package FastRTPS

dirk-thomas:
I am not sure if when / if we want to switch but this allows us to package ROS 2 with FastRTPS.

Ready for review.

### DIFF
--- a/job_templates/packaging_job.xml.template
+++ b/job_templates/packaging_job.xml.template
@@ -31,6 +31,16 @@ To use the default branch on all repositories, use an empty string.</description
           <defaultValue></defaultValue>
         </hudson.model.StringParameterDefinition>
         <hudson.model.ChoiceParameterDefinition>
+          <name>CI_USED_RMW_IMPL</name>
+          <description/>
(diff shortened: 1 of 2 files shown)


job_templates/packaging_job.xml.template
```diff
@@ -31,6 +31,16 @@ To use the default branch on all repositories, use an empty string.</description
           <defaultValue></defaultValue>
         </hudson.model.StringParameterDefinition>
         <hudson.model.ChoiceParameterDefinition>
+          <name>CI_USED_RMW_IMPL</name>
+          <description/>
+          <choices class="java.util.Arrays$ArrayList">
+            <a class="string-array">
+              <string>OpenSplice</string>
+            <s  tring>FastRTPS</string>
```
wjwwood: Looks like there is extra white space in the tag?

wjwwood:
lgtm, other than the comment.

job_templates/packaging_job.xml.template
dirk-thomas: Fixed.

PR #267, opened 2016-09-07
use Kinetic instead if Indigo for the bridge

### LINKED ISSUE #266: ROS 1 bridge not in Linux binary distribution
@geoffviola I reverted your wiki change here:

https://github.com/ros2/ros2/wiki/Linux-Install-Binary/_compare/419c58fb171c6f66baa7d5fd8c861c5afc43b8bb...ae967707b9f9651a88ae377eb412444c1ffadb93

The dynamic_bridge is supposed to be in the Linux binary package, but that broke when we changed to Xenial (kinetic) from Trusty (indigo). Thanks for catching the problem for us though, we're looking into it.

dirk-thomas:
Until now the packaging jobs tried to source indigo which isn't available on Xenial and therefore skipped building the bridge.

Fixes #266.


dirk-thomas:
The latest Linux package has the `dynamic_bridge` binary again.

wjwwood:
+1

PR #283, opened 2016-10-31
new location of rmw fastrtps repo

dirk-thomas:
FYI

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -43,10 +43,6 @@ repositories:
     type: git
     url: https://github.com/eProsima/Fast-RTPS.git
     version: master
-  eProsima/ROS-RMW-Fast-RTPS-cpp:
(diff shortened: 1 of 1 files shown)


ros2.repos
```diff
@@ -99,6 +95,10 @@ repositories:
     type: git
     url: https://github.com/ros2/rmw_connext.git
     version: master
+  ros2/rmw_fastrtps_cpp:
```
wjwwood: I agree, and the repository should probably be called the same, currently it is also `rmw_fastrtps_cpp`: https://github.com/ros2/rmw_fastrtps_cpp

dirk-thomas: Done 8ba5a57cbdd0523bc407b9f7875d44f87b1aae57

PR #299, opened 2016-12-23
use specific FastRTPS version before breaking change

dirk-thomas:
Since the latest changes to FastRTPS break many tests across all platforms (see ros2/rmw_fastrtps#81) this patch locks in a version of FastRTPS before that change for now. We can move back to master if it has been fixed (or we know what we need to change in our code to fix the failing tests).

@richiprosima FYI

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -42,7 +42,7 @@ repositories:
   eProsima/Fast-RTPS:
     type: git
     url: https://github.com/eProsima/Fast-RTPS.git
-    version: master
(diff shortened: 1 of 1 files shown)


dirk-thomas:
We had talked about it yesterday already in the office and in order to not have the jobs fail all the tests for the next days (and being unable to see problem with upcoming PRs) I took the liberty to merge this. I also create #300 which reverts back to using master to track that step.

PR #300, opened 2016-12-23
use FastRTPS from master again

dirk-thomas:
This reverts the change from #299 but requires either changes to FastRTPS before or an update to ROS 2 to address the test failures.


dirk-thomas:
See https://github.com/ros2/rmw_fastrtps/issues/81#issuecomment-269365868

PR #304, opened 2017-01-03
add ament_cmake_ros repo

dirk-thomas:
The new repo [ament_cmake_ros](https://github.com/ros2/ament_cmake_ros) contains the CMake option `BUILD_SHARED_LIBS` which defaults to `ON` (rather than `OFF` which is the CMake default). I think this is a cleaner approach then to modify `ament_tools` as proposed in ament/ament_tools#128 since even when the packages are being built with `cmake && make && make install` they behave the same.

Please review.

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -51,6 +51,10 @@ repositories:
     type: git
     url: https://github.com/ros/geometry2.git
     version: ros2
+  ros2/ament_cmake_ros:
+    type: git
+    url: https://github.com/ros2/ament_cmake_ros.git
+    version: master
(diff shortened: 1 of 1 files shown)


wjwwood:
The package lgtm, but doesn't (most) every package now need to depend on this instead of `ament_cmake`?

dirk-thomas:
Yes, as soon as they want to drop the `SHARED` keyword and rely on `BUILD_SHARED_LIBS`.

wjwwood:
Ok, so we want to merge this before doing each of those. +1

PR #315, opened 2017-03-01
fix key to match actual location of poco_vendor repo

dirk-thomas:


wjwwood:
@ros2/team everyone will need to move or delete the `ros/poco_vendor` folder if you want to update your local tree with the latest `ros2.repos` file. I usually just mv the existing `src` folder to a backup and get everything fresh from the new `ros2.repos` file.

PR #317, opened 2017-03-07
Added two thirdparty libraries for Fast RTPS

richiware:
After accepting eProsima/Fast-RTPS#26, FastRTPS needs two thirdparty libraries: asio and tinyxml2. These libraries were inserted as Git submodules. Ament system doesn't update Git submodules. Alternatively I found that inserting these libraries in ros2.repos file works successfully.

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -43,6 +43,14 @@ repositories:
     type: git
     url: https://github.com/eProsima/Fast-RTPS.git
     version: master
+  eProsima/Fast-RTPS/thirdparty/asio:
+    type: git
+    url: https://github.com/chriskohlhoff/asio.git
+    version: master
+  eProsima/Fast-RTPS/thirdparty/tinyxml2:
+    type: git
+    url: https://github.com/leethomason/tinyxml2.git
+    version: master
   ros/class_loader:
     type: git
     url: https://github.com/ros/class_loader.git


dirk-thomas:
I triggered a set of CI jobs for this:

* http://ci.ros2.org/job/ci_linux/2297/
* http://ci.ros2.org/job/ci_osx/1754/
* http://ci.ros2.org/job/ci_windows/2273/

dirk-thomas:
This approach does not work:

> Could not clone repository 'https://github.com/eProsima/Fast-RTPS.git': fatal: destination path '.' already exists and is not an empty directory.

We will use a previous version of FastRTPS until this is resolved. See #318.

We will follow up on this topic soon to suggest a different way to handle these dependencies. E.g. it should be possible to use the system packages (http://packages.ubuntu.com/source/xenial/asio http://packages.ubuntu.com/source/xenial/tinyxml2) when they are available.

dhood:
you could also just clone the two new repos into a directory that isn't a subdirectory of eProsima/Fast-RTPS, right? we could just import them like this:
```
  eProsima/asio:
    type: git
    url: https://github.com/chriskohlhoff/asio.git
    version: master
  eProsima/tinyxml2:
    type: git
    url: https://github.com/leethomason/tinyxml2.git
    version: master
```

It's not an issue to use a fixed version of fastrtps at the moment, but depending on how long this alternative approach to handle the dependencies takes this might be an appropriate compromise.

dirk-thomas:
@dhood How does FastRTPS find the dependencies in these sibling folders?

dhood:
d'oh, of course. OK new suggestion: `vcs import -w 1`. cloning into the subdirectory can work, as long as fastrtps gets cloned first.

mikaelarguedas:
There are several limitations with this approach:
- FastRTPS doesnt declare dependencies in a manifest so we can't create a
topological order from there
- There are several relative includes in the code, makefiles and cmake
(e.g.
https://github.com/eProsima/Fast-RTPS/blob/4666e34d0f04267e58ff4f13e1b0a356741a3b70/examples/C%2B%2B/DeadlineQoSExample/CMakeLists.txt#L37)
that lock down the destination of these third party packages.

I do agree that having a way to externally know the dependencies of the
package and simply find_package them would be more transparent and
convenient.

dirk-thomas:
That could work (no idea if I wrote vcstool in a way that it processes the repos in order :wink:). But I think we should push eProsima to rework their CMake code to not require those submodules. They can stil do it optionally if the dependency is not findable on the system.

mikaelarguedas:
:+1: That would also allow us to use upstream when available rather than building dependencies that already have binaries. We can then make it work on the platforms that don't have a package for it (e.g. what we do for poco with poco_vendor)

dhood:
> That could work (no idea if I wrote vcstool in a way that it processes the repos in order :wink:). 

It appears so!

> But I think we should push eProsima to rework their CMake code to not require those submodules. They can stil do it optionally if the dependency is not findable on the system.

OK, well we can revisit this suggestion if using a pinned version affects us before other solutions are implemented.

richiware:
eProsima/Fast-RTPS@67e42c8a52ae631ad198f3da20ceb782fbb6f6bf adds two CMake modules: one to find asio and other one to find tinyxml2. Is it useful for you to find a solution?

dirk-thomas:
@richiware Since then we have switched (#318) to an older version of FastRTPS. In order to switch back to using the master branch we would like to see the following changes to FastRTPS:

* When the two dependencies `TinyXML2` and `asio` are already installed on the system (from e.g. Debian packages) they should be used instead of a custom version pulled in via submodules.
* Also `TinyXML2` should not be compiled within FastRTPS in that case (https://github.com/eProsima/Fast-RTPS/blob/ad59a6cf29e6f15a105193ea304cff645f0b9137/src/cpp/CMakeLists.txt#L105). Instead only the provided headers as built libraries should be used.
* The CMake code should contain a call like `find_package(TinyXML2)` in order for our build tool to extract that dependency in case we need a package with the same name to be build on e.g. Windows.

Is that something you could work on rather soon or would accept a PR for since it is prevent us from using newer versions of FastRTPS?

dirk-thomas:
@richiprosima

richiware:
We are testing a new release 1.4.0. Next week I will make the changes you've suggested. Thanks

dirk-thomas:
@richiware Thanks, please comment here once they are available or if you have any questions.

richiprosima:
@dirk-thomas eProsima/Fast-RTPS@d0133e00ab32ef9dc107364393efff201b5a3663 modifies FindAsio and FindTinyXML2. Now it can detect if libraries are installed on system.

dirk-thomas:
@richiprosima I cloned the latest master and invoked `cmake` on it and it failed with the following error for me:

```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
ASIO_INCLUDE_DIR (ADVANCED)
   used as include directory in directory /tmp/ws/src/Fast-RTPS/src/cpp
```

After installing the dependency `libasio-dev` I was able to build FastRTPS successfully. If you are missing `libtinyxml2-dev` the problem is the same. But I guess the CMake logic should be updated to fail with a clear error message.

With that being said it builds for me on Linux. I will check for OS X and Windows next.

BrannonKing:
I prefer using the system packages instead of including all dependencies in my build time. However, it would be nice if VCS supported that idea. You should be able to specify an "apt" or "dpkg" or "dnf" requirement.

dirk-thomas:
Can you please clarify what you mean with this:

> However, it would be nice if VCS supported that idea. You should be able to specify an "apt" or "dpkg" or "dnf" requirement.

BrannonKing:
To clarify, we have this in the ros2.repos file at the moment:
```
  eProsima/Fast-RTPS:
    type: git
    url: https://github.com/eProsima/Fast-RTPS.git
    version: 4ae016fb90ea383e00b956fcba977f486894a0e5
```

Suppose it had this additional information:
```
   requires-apt: libasio-dev, libtinyxml2-dev
   requires-dnf: asio-devel, tinyxml2-devel
   requiers-pacman: ....
   requires-pip3: flake8
```

If any of the required packages are missing, it should exist and give the user a sudo command to run to install the missing packages. It may be that specific versions are needed. In that case, though, the ROS library packager should include that through one of the other VCS package options.

Cartographer's instructions give a list of "apt-get" calls. I don't want to have to send instructions as to what system-level packages need to be pre-existing. I want it to work out of the box.

wjwwood:
@BrannonKing we have a system for that where the packages self-describe what they need in the `package.xml` and then they get installed with `rosdep`. So if we had something what you describe in the repos file (for things that don't have a package.xml) then it would probably be good to use `rosdep` keys instead of putting the apt stuff in directly.

Also it always seems simple to just add some apt things in, but in reality it gets tricky fast when you're modifying the whole systems packages. It's not something that scales without some effort.

dirk-thomas:
`vcstool` which is handling the `.repos` file is only focused on cloning / updating / etc. repositories. The task to install dependencies is addressed by `rosdep` which relies on the package to have a manifest names `package.xml` specifying the dependencies. `FastRTPS` doesn't contain such a manifest therefore `rosdep` can't install these for you at the moment.

The reason that the metainformation is provided by the package itself is because it is also necessary to create binary packages. So I don't think this information belongs in the `.repos` file.

dirk-thomas:
@nuclearsandwich has created Chocolatey packages for these two dependencies and is working on a PR for FastRTPS to use those (PR upcoming). I will close this since it will be replaced by #332.

PR #324, opened 2017-03-22
Move geometry2 (both locally, and on github).

clalancette:
We want to switch to using our fork, instead of a branch on
the original.  This matches what we do for other repositories.

Signed-off-by: Chris Lalancette <clalancette@osrfoundation.org>

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -47,9 +47,9 @@ repositories:
     type: git
     url: https://github.com/ros/class_loader.git
     version: ros2
-  ros/geometry2:
+  ros2/geometry2:
     type: git
-    url: https://github.com/ros/geometry2.git
+    url: https://github.com/ros2/geometry2.git
     version: ros2
   ros2/ament_cmake_ros:
     type: git
--- a/ros2.repos
+++ b/ros2.repos
@@ -47,10 +47,6 @@ repositories:
     type: git
     url: https://github.com/ros/class_loader.git
     version: ros2
-  ros2/geometry2:
-    type: git
-    url: https://github.com/ros2/geometry2.git
-    version: ros2
   ros2/ament_cmake_ros:
     type: git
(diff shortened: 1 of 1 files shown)


ros2.repos
```diff
@@ -47,9 +47,9 @@ repositories:
     type: git
     url: https://github.com/ros/class_loader.git
     version: ros2
-  ros/geometry2:
+  ros2/geometry2:
```
mikaelarguedas: Nit: should be placed alphabetically in the list

clalancette: Whoops.  Done now.

dhood:
do we have a standard yet for how we organise ported ros1-to-ros2 repos in this file? Was `geometry2` in the `ros` directory because it was being pulled in from the `ros` github org, or because it is a ported ros1 package? Maybe we haven't really thought about it before, but at some point we should be decide so we can be consistent so that users know where to look for files

mikaelarguedas:
I think the logic was the folder name matches the org unit it's pulled from. But that's actually a guess based on observations, I don't know if we have a standard for that. It could make sense to dissociate the core of ros2 from the packages ported from ros1, at the folder level and even maybe at the org unit level

dhood:
OK it's fine by me to match the org unit for now and we can always change it in the future if it makes more sense to have a different structure for ported packages :+1:

clalancette:
Thanks all, merged now.  I think this won't happen automatically for people; should I send out a notification somewhere that the location has changed?

mikaelarguedas:
Any change to the repos file will require people to pull the new one and import it in their workspace. You are right that people already having geometry2 in their workspace will have a hiccup of duplicate packages (ament will notify the duplication and the location of the conflicting packages).
@ros2/team please think about removing the current version of geometry2 in the ros subfolder before pulling this updated repos file.

PR #332, opened 2017-04-11
Use Fast-RTPS fork containing patched upstream master.

nuclearsandwich:
This will update ros2.repos to use a temporary fork of Fast-RTPS that incorporates CMake changes needed to use our Chocolatey packages on Windows.

Per @dirk-thomas's suggestion I've updated the repos file to test the changes more widely before we propose them upstream.

This PR as-is depends on https://github.com/ros2/rmw_fastrtps/pull/101 but can be altered to use rmw_fastrtps/try-fastrtps-master if we want to test this before landing that pull request.

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -41,8 +41,8 @@ repositories:
     version: master
   eProsima/Fast-RTPS:
     type: git
-    url: https://github.com/eProsima/Fast-RTPS.git
-    version: 4ae016fb90ea383e00b956fcba977f486894a0e5
+    url: https://github.com/ros2/Fast-RTPS.git
+    version: ros2
   ros/class_loader:
     type: git
     url: https://github.com/ros/class_loader.git


mikaelarguedas:
I think it's ok to merge ros2/rmw_fastrtps#101 before proceeding with this.

Great work! awesome that we can now track Fast-RTPS master again :tada:

nuclearsandwich:
> Great work! awesome that we can now track Fast-RTPS master again 🎉

I should mention that my fork has no "automatic" sync to master, I've just rebased it and it can be re-rebased periodically but we should aim to give this a thorough test and upstream soon so we're no longer lagging by manual intervention.

mikaelarguedas:
I think it's ok because we don't plan on relying on that fork for long :+1:

dirk-thomas:
@nuclearsandwich How woud you like to proceed on this? Have you deployed the Chocolatey packages on all Windows nodes? Afterwards should we start using the fork if FastRTPS or do you want to create a PR against FastRTPS and wait a few days if it gets accepted quickly?

Karsten1987:
once this gets merged, please be aware of updating the windows installation
wiki pages.

nuclearsandwich:
> Have you deployed the Chocolatey packages on all Windows nodes? 
Yep. windshield, portable, eatable, and icecube all have them installed, as does the VM image I'm prepping for internal testing

> How would you like to proceed on this? Afterwards should we start using the fork if FastRTPS or do you want to create a PR against FastRTPS and wait a few days if it gets accepted quickly?

Per our previous offline discussions @dirk-thomas I think it's worthwhile for these CMake changes to see slightly wider testing, especially on non-Windows development machines before submitting them upstream. If @ros2/team uses our fork for the remainder of this week when working and no issues surface, I'll open a PR next week.

> once this gets merged, please be aware of updating the windows installation wiki pages.

This is a good tip, thanks, I'll get an edit queued.

nuclearsandwich:
> LGTM with some CI builds for all platforms.

🍎  [Build Status](http://ci.ros2.org/job/ci_osx/1882/)

🐧 pending available executor

nuclearsandwich:
@mikaelarguedas is [this warning](http://ci.ros2.org/job/ci_osx/1882/warnings10Result/new/NORMAL/) on Mac OS worth blocking this or should we merge it and work it out there?

mikaelarguedas:
given that this warning is not in our code base I think it's not a blocker. A PR fixing it should be open ASAP though so that we can have green builds in a timely manner. @ros2/team anybody think it should be hold because of this warning ?

nuclearsandwich:
Installation instructions on the wiki have been updated.

mikaelarguedas:
Thanks! I could use some of you github knowledge here. do you know why the [revision](https://github.com/ros2/ros2/wiki/Windows-Development-Setup/_compare/9c95bf76295582099cf4cdb2c05c7c3bcc7389c4...4a7c3fbfe5ba0294af57d13b5c463d8044e7612d#diff-ec5890d4bba61a38f6ce3669d34fac21) claims to be touching 12 files ?
Nit: does the order of installation matters ? if not can you list the deps alphabetically rather than adding one at the beginning of the list and the other one at the?
Thanks!

nuclearsandwich:
> Thanks! I could use some of you github knowledge here. do you know why the revision claims to be touching 12 files ?

The revision link you have there is scoped to a single wiki page, so it's only showing that page's changes. But the actual git diff between the two commits, `9c95bf7` (the previous commit touching _that file_) and `4a7c3fb` (my commit) includes other changes made to other files in other commits. A [comparison url](https://github.com/ros2/ros2/wiki/_compare/9c95bf76295582099cf4cdb2c05c7c3bcc7389c4...4a7c3fbfe5ba0294af57d13b5c463d8044e7612d) with the wiki page component removed shows all the other changes.

It looks like they're not scoping the diffstat to the filename. So they're performing  `git diff --stat 9c95bf7..4a7c3fb` rather than `git diff --stat 9c95bf7..4a7c3fb Windows-Development-Setup.md` in both cases.


> Nit: does the order of installation matters ? if not can you list the deps alphabetically rather than adding one at the beginning of the list and the other one at the?

I did alphabetize in a few places, mostly on the command line. I'll fix the actual list though.

mikaelarguedas:
awesome thanks for the explanation!

nuclearsandwich:
> given that this warning is not in our code base I think it's not a blocker. A PR fixing it should be open ASAP though so that we can have green builds in a timely manner. @ros2/team anybody think it should be hold because of this warning ?

Since I have your consent as a reviewer and the reasoning you provided is sound, I am going ot merge this so that it can start getting used more widely.

@ros2/team this update to the repos file will bring in a relatively fresh Fast-RTPS version. Please let me know if the CMake changes cause any trouble for you finding dependencies on other platforms. If we don't notice any show-stoppers I'll propose the changes upstream next week.

mikaelarguedas:
can you please link here to the corresponding PR on Fast-RTPS to fix the warning?
Thanks!

nuclearsandwich:
> can you please link here to the corresponding PR on Fast-RTPS to fix the warning?
> Thanks!

I'm honestly not sure what an acceptable fix is given that this warning only appears on one of hte platforms we ran it against and isn't in our codebase. I'd be happy to put the work in with a little guidance but I've not spent enough time in C++ codebases or ROS 2 / Fast RTPS such that a solution other than forcing the warning off for some platforms is clear to me.

mikaelarguedas:
The warning indicates a class member never being used. If it is a trivial fix (like casting it or removing it) it's ok to open a PR. If we're not sure about the fix we should open an issue on the upstream repo for them to address (with a link to the [culprit commit](https://github.com/eProsima/Fast-RTPS/commit/2d07bdac258f8140ceebca99588bb1c100976cf9#diff-667a9e1219bfacd17199a05bcced6d75R125)). In any case you should not spend much time on it because as you said it's not in our codebase, but we need to make sure there we did what's necessary for this to be fixed in the near future.

wjwwood:
I went ahead and open a pr for this, we can see what they think about it: https://github.com/eProsima/Fast-RTPS/pull/88

PR #341, opened 2017-04-18
track upstream Fast-RTPS

mikaelarguedas:
now that https://github.com/eProsima/Fast-RTPS/pull/91 has been merged no need to work from the fork anymore

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -41,8 +41,8 @@ repositories:
     version: master
   eProsima/Fast-RTPS:
     type: git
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
🎉

dirk-thomas:
@nuclearsandwich Can our fork be removed?

mikaelarguedas:
I think that it's ok to keep it so that we open our PRs from a single place rather than all having our forks under personal org units. I dont feel strongly about it though

Issue #342, opened 2017-04-19
Cartographer support


clalancette:
This is a meta-ticket for Cartographer support in ROS2.

mikaelarguedas:
@clalancette Can this ticket be closed ?

mikaelarguedas:
Yeah I don't think we should keep a ticket open for every package we want to improve. Specific issues about what to improve is more relevant IMO. (it's also to cleanup the waffleboard because my screen is getting too small for this ticket)

mikaelarguedas:
> (like ros2/teleop_twist_joy#3, etc)

This is not related to cartographer and will be useful for any demo using teleop of any robot

clalancette:
Heh, OK.  Closing it out, then.

YuehChuan:
HI, wonder is there some details  about how to  make ROS2  on Cartographer? TQ!

clalancette:
@YuehChuan We're still in the process of writing the documentation for it.  Once that is done, it should be available in the https://github.com/ros2/turtlebot2_demo repository (or, at least, linked to from there).

mikaelarguedas:
Hi @YuehChuan, we will give instructions of how to build and run the cartographer demo on [this repository](https://github.com/ros2/turtlebot2_demo). The instructions will be available in the next week or 2

YuehChuan:
@clalancette , @mikaelarguedas ,Thanks for the info! I just notice the reply now. Now I'm compiling ros2 beta2 on raspberry pi3.

PR #345, opened 2017-04-25
Add vision_opencv to the list of repositories.

clalancette:
Also do a minor bit of rearrangement for an alphabetical
listing.

connects to ros2/ros2#342

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -47,13 +47,13 @@ repositories:
     type: git
     url: https://github.com/ros/class_loader.git
     version: ros2
+  ros/urdfdom_headers:
+    type: git
+    url: https://github.com/ros2/urdfdom_headers.git
+    version: math_h
   ros2/ament_cmake_ros:
     type: git
     url: https://github.com/ros2/ament_cmake_ros.git
-    version: master
-  ros2/rcutils:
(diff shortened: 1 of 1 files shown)


ros2.repos
```diff
@@ -47,14 +47,14 @@ repositories:
     type: git
     url: https://github.com/ros/class_loader.git
     version: ros2
+  ros/urdfdom_headers:
```
mikaelarguedas: I think that the place in the file was good, it should be cloned in the ros2 folder though, like the other forked repositories.

mikaelarguedas:
This depends on https://github.com/ros2/vision_opencv/pull/1 and should thus be merged after

ros2.repos
clalancette: Oh, good point.  I missed that it was forked, and not the original.  I'll move it back and put it in the ros2 subfolder.

clalancette:
Yeah, definitely.  I'm fairly confident in that one now (I'm just going to wait until it finishes the last CI), so I figured I would get this queued up.  Thanks for looking :).

clalancette:
All right, I merged up ros2/vision_opencv, so I'll merge this up now.

PR #351, opened 2017-05-11
add issue template

mikaelarguedas:
This adds a default issue template to the ros2/ros2 repo. To see what it looks like see https://github.com/mikaelarguedas/ros2/issues/new.
Looking for feedback on the content of the template and to know is we should propagate it to all repos in the ros2 organization.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -5,13 +5,18 @@ For Bug report or feature requests, please fill out the relevant category below
 
 ## Bug report
 
-| Required Info | |
-|---|---|
-|Operating System|<!-- OS and version (e.g. Windows 10, Ubuntu 16.04...) -->|
-|Installation type|<!-- binaries or from source  -->|
-|Version or commit hash|<!-- Output of git rev-parse HEAD, release version or repos file  -->|
-|DDS implementation|<!-- rmw_implementation used (e.g. Fast-RTPS, RTI Connext, etc -->|
-|Client library (if applicable)|<!-- e.g. rclcpp, rclpy or N/A -->|
+**Required Info:**
+
+- Operating System:
+  - <!-- OS and version (e.g. Windows 10, Ubuntu 16.04...) -->
+- Installation type:
+  - <!-- binaries or from source  -->
+- Version or commit hash:
+  - <!-- Output of git rev-parse HEAD, release version, or repos file  -->
+- DDS implementation:
+  - <!-- rmw_implementation used (e.g. Fast-RTPS, RTI Connext, etc -->
+- Client library (if applicable):
+  - <!-- e.g. rclcpp, rclpy, or N/A -->
 
 #### Steps to reproduce issue
 <!-- Detailed instructions on how to reliably reproduce this issue http://sscce.org/
@@ -26,7 +31,8 @@ For Bug report or feature requests, please fill out the relevant category below
 
 #### Additional information
 
-
+<!-- If you are reporting a bug delete everything below
+     If you are requesting a feature deleted everything above this line -->
 ----
 ## Feature request
 
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -31,7 +31,7 @@ For Bug report or feature requests, please fill out the relevant category below
 ## Feature request
 
 #### Feature description
-<!-- Description in a few sentences what the feature consists in and what problem it will solve -->
+<!-- Description in a few sentences what the feature consists of and what problem it will solve -->
 
 #### Implementation considerations
-<!-- Relevant information of how the feature could be implemented and pros and cons of the different solutions -->
+<!-- Relevant information on how the feature could be implemented and pros and cons of the different solutions -->
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,43 @@
+<!--
+For general questions, please post on discourse: https://discourse.ros.org/c/ng-ros
+For Bug report or feature requests, please fill out the relevant category below
+-->
+
+<!-- BUG REPORT -->
+## Bug report
+
+| Required Info | |
+|---|---|
+|Operating System|<!-- OS X/macOS, Ubuntu, or Windows, as well as version (e.g. Windows 10, Ubuntu 16.04) -->|
+|Installation type|<!-- binaries or from source  -->|
+|Version or commit hash|<!-- Output of git rev-parse HEAD, release version or repos file  -->|
+|DDS implementation|<!-- rmw_implementation use (e.g. Fast-RTPS, RTI Connext, etc -->|
+|Client library (if applicable)|<!-- e.g. rclcpp, rclpy or N/A -->|
+
+#### Steps to reproduce issue
+<!-- Detailed instructions on how to reliably reproduce this issue http://sscce.org/
+``` code that can be copy-pasted is preferred ``` -->
+```  ```
+
+#### Expected behavior
+<!-- Action or result expected from performing the steps above -->
+
+#### Actual behavior
+<!-- Action or result observed after performing the steps above -->
+
+#### Contributors with relevant knowledge or expertise (optional)
+<!-- @-mention individuals with knowledge in this problem domain -->
+
+#### Additional information
+<!-- If you want to give additional infromation describing the issue -->
+
+
+----
+<!-- FEATURE REQUEST -->
+## Feature request
+
+#### Feature description
+<!-- Description in a few sentences what the feature consists in and what problem it will solve -->
+
+#### Implementation considerations
+<!-- Relevant information of how the feature could be implemented and pros and cons of the different solutions -->
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -3,37 +3,31 @@ For general questions, please post on discourse: https://discourse.ros.org/c/ng-
 For Bug report or feature requests, please fill out the relevant category below
 -->
 
-<!-- BUG REPORT -->
 ## Bug report
 
 | Required Info | |
 |---|---|
-|Operating System|<!-- OS X/macOS, Ubuntu, or Windows, as well as version (e.g. Windows 10, Ubuntu 16.04) -->|
+|Operating System|<!-- OS and version (e.g. Windows 10, Ubuntu 16.04...) -->|
 |Installation type|<!-- binaries or from source  -->|
 |Version or commit hash|<!-- Output of git rev-parse HEAD, release version or repos file  -->|
-|DDS implementation|<!-- rmw_implementation use (e.g. Fast-RTPS, RTI Connext, etc -->|
+|DDS implementation|<!-- rmw_implementation used (e.g. Fast-RTPS, RTI Connext, etc -->|
 |Client library (if applicable)|<!-- e.g. rclcpp, rclpy or N/A -->|
 
 #### Steps to reproduce issue
 <!-- Detailed instructions on how to reliably reproduce this issue http://sscce.org/
 ``` code that can be copy-pasted is preferred ``` -->
-```  ```
+```
+
+```
 
 #### Expected behavior
-<!-- Action or result expected from performing the steps above -->
 
 #### Actual behavior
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,43 @@
+<!--
+For general questions, please post on discourse: https://discourse.ros.org/c/ng-ros
+For Bug report or feature requests, please fill out the relevant category below
+-->
+
+<!-- BUG REPORT -->
```
clalancette: Here (and elsewhere), I think it would be best to get rid of the comments and just have the categories.  That is because I think it is too busy with all of the comments inline, and the comments are (mostly) repeating what the categories are asking for.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,43 @@
...
+|DDS implementation|<!-- rmw_implementation use (e.g. Fast-RTPS, RTI Connext, etc -->|
+|Client library (if applicable)|<!-- e.g. rclcpp, rclpy or N/A -->|
+
+#### Steps to reproduce issue
+<!-- Detailed instructions on how to reliably reproduce this issue http://sscce.org/
+``` code that can be copy-pasted is preferred ``` -->
+```  ```
+
+#### Expected behavior
+<!-- Action or result expected from performing the steps above -->
```
clalancette: Remove the comment

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,43 @@
...
+
+#### Actual behavior
+<!-- Action or result observed after performing the steps above -->
```
clalancette: Remove the comment.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,43 @@
...
+
+#### Contributors with relevant knowledge or expertise (optional)
+<!-- @-mention individuals with knowledge in this problem domain -->
```
clalancette: Remove the comment.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,43 @@
...
+
+#### Additional information
+<!-- If you want to give additional infromation describing the issue -->
```
clalancette: Remove the comment.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,43 @@
...
+
+
+----
+<!-- FEATURE REQUEST -->
```
clalancette: Remove the comment.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,43 @@
...
+## Feature request
+
+#### Feature description
+<!-- Description in a few sentences what the feature consists in and what problem it will solve -->
```
clalancette: Remove the comment.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,43 @@
...
+## Bug report
+
+| Required Info | |
+|---|---|
+|Operating System|<!-- OS X/macOS, Ubuntu, or Windows, as well as version (e.g. Windows 10, Ubuntu 16.04) -->|
+|Installation type|<!-- binaries or from source  -->|
+|Version or commit hash|<!-- Output of git rev-parse HEAD, release version or repos file  -->|
+|DDS implementation|<!-- rmw_implementation use (e.g. Fast-RTPS, RTI Connext, etc -->|
+|Client library (if applicable)|<!-- e.g. rclcpp, rclpy or N/A -->|
+
```
clalancette: This table is nice, but again seems pretty busy.  Either leave it blank, or maybe have pre-populated lists for each of them?

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,43 @@
...
+<!-- BUG REPORT -->
+## Bug report
+
+| Required Info | |
+|---|---|
+|Operating System|<!-- OS X/macOS, Ubuntu, or Windows, as well as version (e.g. Windows 10, Ubuntu 16.04) -->|
+|Installation type|<!-- binaries or from source  -->|
+|Version or commit hash|<!-- Output of git rev-parse HEAD, release version or repos file  -->|
+|DDS implementation|<!-- rmw_implementation use (e.g. Fast-RTPS, RTI Connext, etc -->|
```
clalancette: use -> used

.github/ISSUE_TEMPLATE.md
clalancette: (though if you leave it, infromation -> information)

.github/ISSUE_TEMPLATE.md
dirk-thomas: @clalancette Please consider editing such changes directly (rather than commenting for them)

.github/ISSUE_TEMPLATE.md
dirk-thomas: I would suggest to remove this subsection all together. I wouldn't ask the reporter of a ticket to come up with this information.

nuclearsandwich:
> Looking for feedback on the content of the template and to know is we should propagate it to all repos in the ros2 organization.

If and when it comes time to propagate the templates we should do so with a script rather than by hand. Should be about a lunchbreak's worth of work to spike something out.

dirk-thomas:
> If and when it comes time to propagate the templates we should do so with a script rather than by hand. 

You can also just clone the ROS 2 repos locally, copy the new file into all repos, use `vcs custom` to add the new file, commit the change and push the new state back. I would think that would certainly be quicker than using the GitHub API.

.github/ISSUE_TEMPLATE.md
mikaelarguedas: fair enough

.github/ISSUE_TEMPLATE.md
mikaelarguedas: How would you suggest doing a pre-populated list ?

.github/ISSUE_TEMPLATE.md
mikaelarguedas: I'll remove the comments where it's redundant, this comment applies to all the `Remove the comment` in this review

.github/ISSUE_TEMPLATE.md
clalancette: I was just thinking about getting rid of the comments, and reducing the verbiage.  For instance, for Operating System, do something more like:

|Operating System|OSX or Ubuntu or Windows or Other|

This will sort of have the problem that people won't necessarily put in the version, so maybe it could be expanded a bit more.

.github/ISSUE_TEMPLATE.md
mikaelarguedas: :+1:

.github/ISSUE_TEMPLATE.md
mikaelarguedas: I think it's better commented out given that as a user I prefer add only relevant information rather than having to "delete everything that doesn't apply"

.github/ISSUE_TEMPLATE.md
mikaelarguedas: @clalancette Please avoid copy paste review comments, the "Here (and elsewhere)" from https://github.com/ros2/ros2/pull/351#discussion_r116773725 is enough

mikaelarguedas:
> You can also just clone the ROS 2 repos locally, copy the new file into all repos, use vcs custom to add the new file, commit the change and push the new state back. I would think that would certainly be quicker than using the GitHub API.

Agreed that was my plan, if we decide that this template makes sense in every repo, which is questionable (e.g. an issue opened on `rmw_*` doesn't need client library or middleware implementation information)

mikaelarguedas:
I removed redundant comments and kept those that I think add context / are useful. Looking for feedback on the table at the beginning: https://github.com/ros2/ros2/pull/351#discussion_r116774400

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,37 @@
...
+
+#### Additional information
+
+
+----
+## Feature request
+
+#### Feature description
+<!-- Description in a few sentences what the feature consists in and what problem it will solve -->
+
```
clalancette: Nitpick: "what the feature consists of"

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,37 @@
...
+#### Implementation considerations
+<!-- Relevant information of how the feature could be implemented and pros and cons of the different solutions -->
```
clalancette: Nitpick: "Relevant information on how the feature..."

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,37 @@
...
+#### Additional information
+
+
+----
+## Feature request
+
+#### Feature description
+<!-- Description in a few sentences what the feature consists in and what problem it will solve -->
+
+#### Implementation considerations
```
clalancette: Might want to add "#### Implementation considerations (optional)"

.github/ISSUE_TEMPLATE.md
mikaelarguedas: Thanks, Can you please commit language nitpicks and typos directly ?

.github/ISSUE_TEMPLATE.md
mikaelarguedas: same as https://github.com/ros2/ros2/pull/351/files#r117349669

.github/ISSUE_TEMPLATE.md
mikaelarguedas: I think in that case we would need to do the same for `#### Additional information`
Overall I think that if people don't know what to put they would put nothing. Looking at other projects like docker or atom they don't add an `(optional)` assuming that people will fill it out if they have something to add. I'm not against it though, I'm just not convinced it's needed

clalancette: I don't really feel strongly about it, so if you don't think it is necessary, that's fine with me.

mikaelarguedas:
@clalancette FYI https://github.com/ros2/ros2/pull/351/commits/7eb2d46fb590d6decc780f4aa6bd26c2b31aa30b

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Bug report
+
+| Required Info | |
+|---|---|
+|Operating System|<!-- OS and version (e.g. Windows 10, Ubuntu 16.04...) -->|
+|Installation type|<!-- binaries or from source  -->|
+|Version or commit hash|<!-- Output of git rev-parse HEAD, release version or repos file  -->|
+|DDS implementation|<!-- rmw_implementation used (e.g. Fast-RTPS, RTI Connext, etc -->|
+|Client library (if applicable)|<!-- e.g. rclcpp, rclpy or N/A -->|
```
wjwwood: For me, a table is very hard to read and edit in raw markdown. What do you think about a list instead?

```
**Required Info:**

- Operating System:
  - <!-- OS and version (e.g. Windows 10, Ubuntu 16.04...) -->
- Installation type:
  - <!-- binaries or from source  -->
- Version or commit hash:
  - <!-- Output of git rev-parse HEAD, release version, or repos file  -->
- DDS implementation:
  - <!-- rmw_implementation used (e.g. Fast-RTPS, RTI Connext, etc -->
- Client library (if applicable):
  - <!-- e.g. rclcpp, rclpy, or N/A -->
```

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,37 @@
...
+
+#### Expected behavior
+
+#### Actual behavior
+
+#### Additional information
+
+
+----
+## Feature request
```
wjwwood: I would add a comment here saying that they should delete this section (everything below the `----`) if they are reporting a bug, or everything above it if they are reporting a feature. It might be obvious, but maybe a reminder will help some people remember to do it.

mikaelarguedas:
Thanks all for the reviews. I tried to address all the comments (updated version available here https://github.com/mikaelarguedas/ros2/issues/new). I'll merge this on Wednesday if I don't get any more feedback by then

PR #358, opened 2017-06-12
replace cli_tools with ros2cli

dirk-thomas:
Depends on ros2/ros2cli#1.

Since the repo `cli_tools` hasn't been part of the beta 1 release and is therefore not referenced anywhere I suggest to delete the repo once this PR has been merged.


dirk-thomas:
Ready for review.

Issue #310, opened 2017-02-03
console_bridge library not found when chaining workspace using binaries


wjwwood:
When using the binaries we produce for Linux and when building a workspace chained off the workspace provided by those binaries, the `components` package (and perhaps others) fail to find `console_bridge`'s library generating this CMake error:

```
CMake Error at /home/william/testing/nightly/ros2-linux/lib/console_bridge/cmake/console_bridge-config.cmake:15 (message):
  Library 'console_bridge' in package console_bridge is not installed
  properly
Call Stack (most recent call first):
  /home/william/testing/nightly/ros2-linux/share/class_loader/cmake/ament_cmake_export_dependencies-extras.cmake:21 (find_package)
  /home/william/testing/nightly/ros2-linux/share/class_loader/cmake/class_loaderConfig.cmake:30 (include)
  CMakeLists.txt:11 (find_package)

```

I used the most recent binary package from our packaging job ([job-#569](http://ci.ros2.org/view/packaging/job/packaging_linux/569/)) on the build farm to reproduce this, but it was originally reported in https://github.com/ros2/ros2/issues/309 with the binaries for `release-beta1`. These are the steps to reproduce the issue:

- `mkdir -p /tmp/nightly/demos_ws/src`
- `cd /tmp/nightly`
- `wget http://ci.ros2.org/view/packaging/job/packaging_linux/lastSuccessfulBuild/artifact/ws/ros2-package-linux.tar.bz2`
- extract `ros2-package-linux.tar.bz2` to the current directory
- `source ./ros2-linux/setup.zsh`
- `cd ./demos_ws/src`
- `git clone https://github.com/ros2/demos.git`
- `cd ..`
- `ament build`

At this point it fails to build with the above error. I also tried with a different build type:

- `rm -rf build`
- `ament build --cmake-args -DCMAKE_BUILD_TYPE=Release --`

This also fails.

wjwwood:
Thanks @mkhansen-intel for the original report.

mikaelarguedas:
validated that https://github.com/ros/console_bridge/pull/45 fixes this problem. But it hasn't been merged yet

PR #373, opened 2017-06-28
use fork of console bridge with relocatable cmake files

mikaelarguedas:
This is needed for people to be able to build overlay ws on top of our packaging artifacts (see https://github.com/ros2/ros2/issues/310). This can be reverted once https://github.com/ros/console_bridge/pull/45 has been merged
@nuclearsandwich FYI

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -197,5 +197,5 @@ repositories:
     version: ros2
   vendor/console_bridge:
     type: git
-    url: https://github.com/ros/console_bridge.git
-    version: master
+    url: https://github.com/nuclearsandwich/console_bridge.git
+    version: ros2-beta2
--- a/ros2.repos
+++ b/ros2.repos
@@ -197,5 +197,5 @@ repositories:
     version: ros2
   vendor/console_bridge:
     type: git
-    url: https://github.com/nuclearsandwich/console_bridge.git
-    version: ros2-beta2
+    url: https://github.com/ros2/console_bridge.git
+    version: ros2


nuclearsandwich:
Rather than use a fork under my account, I'd like to either move the branch or transfer my entire fork to the ROS 2 org. Changing the branch location won't affect CI so I think it's safe to run as it is now but I'd like to be using a branch of ros2/console_bridge or ros/console_bridge so that other maintainers have reliable edit capabilities.

mikaelarguedas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2835/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/339/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2291/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2971/)

* Linux [Build Status](http://ci.ros2.org/job/ci_turtlebot-demo_linux/64/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_turtlebot-demo_linux-aarch64/78/)

mikaelarguedas:
> Rather than use a fork under my account, I'd like to either move the branch or transfer my entire fork to the ROS 2 org. 

Sounds good to me. I launched ci on your branch for now to see if it's safe to track that branch

nuclearsandwich:
To my unfamiliar eyes the Windows test failures look like the usual flakey suspects but given the relative stability of recent Windows builds I'm worried there may be actual issues there.

mikaelarguedas:
yeah I'm a bit concerned too because the other builds on master didnt show this failures (and they are not usually flaky), I'll retrigger one just to have another data point

mikaelarguedas:
Windows round 2 to confirm if this introduced failures
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2975/)

mikaelarguedas:
@nuclearsandwich hmm so the same tests failed on 2 different machines with this patch while every other job ran in between using master passed. So I think there is a problem on this branch to investigate

nuclearsandwich:
Another option would be to see if https://github.com/ros/console_bridge/pull/45's changes without mine pass CI. Using the GNU InstallDirs version of console_bridge will break some downstream builds until I add the gnuinstalldirs path to dh_shlibdeps for the downstreams that need it but that's a known fix at this point whereas it wasn't when I reverted the support. There is one [other change in my branch](https://github.com/ros/console_bridge/pull/46/files#diff-af3b638bc2a3e6c650974192a53c7291R16) which deals with where the CMake config files are placed. I don't know where the cmake files are placed when using gnuinstalldirs or whether they're somewhere they can be found when building on top of the packaged console_bridge. If there's a reasonable chance that they go somewhere usable we may get away with just targeting https://github.com/ros/console_bridge/pull/45 directly (assuming of course that the errors are introduced by my changes and not https://github.com/ros/console_bridge/pull/45).

Lastly, when not using GNU InstallDirs it sets the cmake datarootdir to share so that cmake files appear alongside other ROS 2 cmake files in /opt/ros/r2b2/share/cmake/PACKAGENAME

mikaelarguedas:
sorry @nuclearsandwich  I missed that notif, here is CI for  ros/console_bridge#45 only
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2846/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/350/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2302/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2985/)
* Linux [Build Status](http://ci.ros2.org/job/ci_turtlebot-demo_linux/68/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_turtlebot-demo_linux-aarch64/83/)

nuclearsandwich:
I've set up a [ros2 fork](https://github.com/ros2/console_bridge) of console_bridge that includes the upstream PR 45 and am building that on the farm now.

Some downstream packages may fail and that's expected until I patch them for the multiarch dirs. I'll try to figure out which ones will need it and preempt them.

As well I would like to make sure that the cmake configuration files are found and can be used by overlay workspaces building downstream packages from source. @dirk-thomas is going to test that once the console_bridge package is ready.

PR #386, opened 2017-08-09
remove obsolete repo ament_python

dirk-thomas:
Connect to ament/ament_python#3.


nuclearsandwich:
🎊 one less thing to package!

Issue #306, opened 2017-01-12
Control shared/static linking via BUILD_SHARED_LIBS and ament_cmake_ros


esteve:
Just an issue to aggregate all the connected PRs so that it shows up on the waffle board

dirk-thomas:
The following CI jobs cover:

* ament/gtest_vendor#4
* ament/gmock_vendor#3
* ros2/rmw#98
* ros2/rcl#115

---

* [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_linux&build=2415)](http://ci.ros2.org/job/ci_linux/2415/)
* [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_osx&build=1868)](http://ci.ros2.org/job/ci_osx/1868/)
* [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_windows&build=2536)](http://ci.ros2.org/job/ci_windows/2536/)

dirk-thomas:
@esteve Waiting for updates / responses on the following PRs:

* ros2/rosidl#183
* ros2/rosidl_typesupport#10
* ros2/rmw_fastrtps#87

dirk-thomas:
* [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_linux&build=2457)](http://ci.ros2.org/job/ci_linux/2457/)
* [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_osx&build=1910)](http://ci.ros2.org/job/ci_osx/1910/)
* [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_windows&build=2575)](http://ci.ros2.org/job/ci_windows/2575/)

esteve:
@dirk-thomas I assume you didn't pass `BUILD_SHARED_LIBS` to the CI build (i.e. the build is using the default `SHARED` value), right? If that's the case, it seems the code in https://github.com/ros2/rosidl_typesupport/pull/10 doesn't read the `BUILD_SHARED_LIBS` variable as expected. Should `ament_cmake_ros` be added as an export dependency and added to https://github.com/ros2/rosidl_typesupport/blob/master/rosidl_typesupport_c/rosidl_typesupport_c-extras.cmake.in and https://github.com/ros2/rosidl_typesupport/blob/master/rosidl_typesupport_cpp/rosidl_typesupport_cpp-extras.cmake.in ?

dirk-thomas:
[...]

PR #412, opened 2017-10-12
aarch64 cross compilation

lmayencourt:
This pull request will allow the cross-compilation of ROS2 for aarch64. 

### Setup the development environment
The same normal setup is required for the cross-compilation -> https://github.com/ros2/ros2/wiki/Linux-Development-Setup

### Get a aarch64 toolchain 
We recommend to use the toolchain provided by Linaro (https://releases.linaro.org/components/toolchain/binaries/latest-5/aarch64-linux-gnu/)
`wget https://releases.linaro.org/components/toolchain/binaries/latest-5/aarch64-linux-gnu/gcc-linaro-5.4.1-2017.05-x86_64_aarch64-linux-gnu.tar.xz`
`tar -xf gcc-linaro-5.4.1-2017.05-x86_64_aarch64-linux-gnu.tar.xz`

### Get the source
Create a workspace and clone all repos:
`git clone https://github.com/ros2/ros2.git -b aarch64_cross_compilation`
`cd ros2/arm_crosscompilation`
`mkdir src`
`vcs-import src < ../ros2.repos`

### Export the CC toolchain
Set the toolchain available for CMake:
`` export CROSS_COMPILE=`pwd`/../../gcc-linaro-5.4.1-2017.05-x86_64_aarch64-linux-gnu/bin/aarch64-linux-gnu- ``

### Remove python support
No target filesystem with python libraries is provided here, so python is not supported (only cpp). A few packages must be modified to allow the compilation of ROS2:
- remove rosidl python in rosidl_default_generators
-- src/ros2/rosidl_typesupport/rosidl_default_generators/CMakeLists.txt: comment line 11 with a `#`
-- src/ros2/rosidl_typesupport/rosidl_default_generators/package.xml: comment line 16 with `<!-- ... -->`
- ignore examples/rclpy
-- src/ros2/examples/rclpy : add an empty file called AMENT_IGNORE
- ignore test_msgs (this one use the path to the python libraries during the install step and in this case try to install under /test_msgs because of the missing path !)
-- src/ros2/rcl_interfaces/test_msgs : add an empty file called AMENT_IGNORE 
- ignore ...

You can do all of this with: 
`sed -e '/py/ s/^#*/#/' -i src/ros2/rosidl_typesupport/rosidl_default_generators/CMakeLists.txt`
`sed -i -r 's/<build(.+?py.+?)/<\!\-\-build\1\-\->/' src/ros2/rosidl_typesupport/rosidl_default_generators/package.xml`
`touch src/ros2/geometry2/AMENT_IGNORE src/ros2/demos/AMENT_IGNORE src/ros2/orocos_kinematics_dynamics/AMENT_IGNORE src/ros2/rclpy/AMENT_IGNORE src/ros2/robot_model/AMENT_IGNORE src/ros2/robot_state_publisher/AMENT_IGNORE src/ros2/system_tests/AMENT_IGNORE src/ros2/urdfdom/AMENT_IGNORE src/ros2/urdfdom_headers/AMENT_IGNORE src/ros2/vision_opencv/AMENT_IGNORE src/ros2/examples/rclpy/AMENT_IGNORE src/ros2/rcl_interfaces/test_msgs/AMENT_IGNORE src/ros2/ros1_bridge/AMENT_IGNORE`

### Trigger a build 
`` src/ament/ament_tools/scripts/ament.py build --force-cmake-configure --cmake-args -DCMAKE_TOOLCHAIN_FILE=`pwd`/aarch64_toolchainfile.cmake -DTHIRDPARTY=ON ``


tfoote:
Thank you for submitting this the toolchain information looks great.

However duplicating our vcs repos file as a repos manifest file is just going to make things more complicated for everyone and from a quick read through it doesn't provide anything that vcs doesn't provide, nor does it support non-git repos. Although we primarily use git we don't require that everyone in the community use git. Could you switch back to using the repos file so that it's easy to compare with our standard development repos file. 

Also the toolchain file is aarch64 specific so should have that in the filename in case we have a different target arch toolchain it won't collide.

Once this is merged the above instructions could be added to the Linux Development wiki in a section at the bottom about cross compiling.

mikaelarguedas:
thanks @lmayencourt for iterating on this. Cool stuff.

Would you mind placing these files in a subfolder (maybe `arm_crosscompilation`?) so that users don't get confused thinking that only this subset of packages works on ARM processors?

Regarding the instructions
- they're very clear thanks :+1: 
- I see that the instructions indicate to build the workspace with `-DTHIRDPARTY=ON`, is it because the .deb packages for Fast-RTPS dependencies are not available for your architecture/platform? (We try to use .deb wherever we can so that the users build only what they can't install from binaries)
- Could you provide an example of targeted platforms / OS that this toolchain and set of packages have been tested on ?

lmayencourt:
Hi @mikaelarguedas, @tfoote and thanks for the comments. 

> Would you mind placing these files in a subfolder (maybe arm_crosscompilation?) so that users don't get confused thinking that only this subset of packages works on ARM processors?

Yes it makes sense ! I will put every thing under `arm_crosscompilation`. 

This setup was tested on Hikey960 and Arm Juno with Ubuntu 16.04 and Busybox. `DTHIRDPARTY=ON` is required because we are not providing any filesystem to the cross-compilation toolchain. It also allows you to create a minimal setup with Busybox. 
But if you have a local copy of the target filesystem with the required dependencies, you can specify its path in the `aarch64_toolchainfile.cmake` and  remove `-DTHIRDPARTRY=ON`

arm_crosscompilation/ros2_arm.repos
```diff
@@ -0,0 +1,193 @@
...
+#    version: ros2
+#  ros2/robot_state_publisher:
+#    type: git
+#    url: https://github.com/ros2/robot_state_publisher.git
+#    version: ros2
+  ros2/ros1_bridge:
+    type: git
+    url: https://github.com/ros2/ros1_bridge.git
+    version: master
+  ros2/ros2cli:
```
mikaelarguedas: This repository contains only python packages, that depend on rclpy. Should this repository be commented out like `rclpy`?

dirk-thomas:
The comment from @tfoote not to duplicate the repos file still applies. While the new file uses the same format as the `ros2.repos` file it still duplicates the information. Maybe the instructions for cross compiling should instead mention which repos can be removed after cloning them or a new command line option could be added to `vcs import` to skip specific entries from the `.repos` file.

tfoote:
Another simple way to knock out the python packages would be to have a command line invocation that uses touch to add AMENT_IGNORE into each of the packages that cannot be crosscompiled. Something like:

```
touch `pwd`/ros2/rclpy/AMENT_IGNORE `pwd`/ros2/orocos_kinematics_dynamics/AMENT_IGNORE
```

This would prevent duplication of the information. And new packages added to the repos file would automatically appear in the cross compile workspace, and could easily be removed if incompatible by extending the list, or pruned once the package is made compatible.

esteve:
@lmayencourt I've crosscompiled ROS2 for the Hikey 960, as well as other ARM boards and [Android](https://github.com/esteve/ros2_java#android) and [iPhone](https://github.com/esteve/ros2_objc#ios), so I'm going to add my two cents, hoping it might be useful to someone else:

- For the aarch64 toolchain, if you are using Debian or Ubuntu, you can just install the toolchain from the repositories (e.g. `$ sudo apt install gcc-aarch64-linux-gnu`)
- The Python 3 interpreter can be crosscompiled in case the distribution of the destination board does have it already. It first needs to be compiled for the host computer, and the use it to bootrstrap to build the interpreter for the target board.
- I started https://github.com/ros2/ros2/issues/306 to enable building ROS2 statically, which is really useful for embedded boards. The last missing piece was agreeing how to build the messages statically. Maybe that can revived if there's any interest.

@lmayencourt let me know if there's anything I can help, I haven't put all this on the wiki, I wasn't sure if anyone would be interested in it, but if there's, I can give it a try.

lmayencourt:
@dirk-thomas @tfoote So if I resume the main idea is to use the `ros2.repos` file to fetch the code and to provide a way (instructions/script/vcs-import new cmd line option/...) to make the required modifications. 
That mean this pull request will only add the `aarch64_toolchainfile.cmake`. 

@esteve thanks for your inputs ! 
One of the goal here is to lay the first stone for a ROS2 on Arm Cortex-M by enabling cross-compilation. So static linking is also a required step. 
I was able to statically link the "publisher" demo by following your work #306

dirk-thomas:
> So if I resume the main idea is to use the ros2.repos file to fetch the code and to provide a way (instructions/script/vcs-import new cmd line option/...) to make the required modifications.
That mean this pull request will only add the aarch64_toolchainfile.cmake.

Yes, plus some description on how to use the toolchain file.

dirk-thomas:
> Yes, plus some description on how to use the toolchain file.

Can you please add the description of how to use the toolchain file (from the first comment in this thread) to a README file in the directory which contains the toolchain file.

Out of curiosity: do you need any further patches at the moment in the other repos in order to cross compile with static libraries for your target?

lmayencourt:
Here we are ! 
Sorry I understood the instructions would be added in the wiki after this PR. My bad.

> Out of curiosity: do you need any further patches at the moment in the other repos in order to cross compile with static libraries for your target?
yes, as @esteve mentioned, other repos must be modified to allow static linking : 
- fastRTPS
- rcutils
- rosidl
- rosidl_typesupport
- rclcpp
- examples 
The patches are mainly removing the `SHARED` in the CMakeList.txt and using `ament_cmake_ros` instead of `ament_cmake` in the package.xml

dirk-thomas:
> yes, as @esteve mentioned, other repos must be modified to allow static linking

I see, so currently you are using shared libraries. Let us know if you need any help moving to static libraries in the future.

arm_crosscompilation/README.md
```diff
@@ -0,0 +1,60 @@
...
+## Trigger a build
+```
+src/ament/ament_tools/scripts/ament.py build --force-cmake-configure --cmake-args -DCMAKE_TOOLCHAIN_FILE=`pwd`/aarch64_toolchainfile.cmake -DTHIRDPARTY=ON
+```
+
+## Installation
+Once the compilation is done you will have to move the generated libraries (`install/lib`) in your target filesystem.
+The default path is under `/lib` but you can also choose another directory and set the `LD_LIBRARY_PATH` variable to it.
+
+*Copyright (c) 2017, ARM Limited and Contributors. All rights reserved.*
```
dirk-thomas: Is this really necessary for these instructions?

esteve:
@lmayencourt I forgot to publish the steps I use for crosscompiling ROS2 for the Raspberry Pi, I finally released the scripts I use and wrote a small README [1][2] The scripts use the compiler toolchain that comes with Debian and Ubuntu, instead of fetching a separate one, and also use Polly [3], instead of a custom CMake toolchain. Polly already comes with CMake toolchains for a ton of targets, it'd be really cool to contribute a toolchain for aarch64 too.

1 - https://discourse.ros.org/t/crosscompiling-ros2-for-raspberry-pi-or-pretty-much-any-other-arm-based-board/2995
2 - https://github.com/esteve/ros2_raspbian_tools
3 - https://github.com/ruslo/polly

esteve:
> @dirk-thomas I see, so currently you are using shared libraries. Let us know if you need any help moving to static libraries in the future.

I'll take you up on that on behalf of @lmayencourt :smile: The last part that was missing for being able to build ROS2 statically (at least up to `rcl`) was https://github.com/ros2/rosidl/pull/183 I proposed adding `ament_ros_cmake` to every message package, but that didn't seem reach a consensus. Is there any other approach you think that might work?

dirk-thomas:
I think the discussion in ros2/rosidl#183 summarizes already what the approach should be. The only thing missing is the behavior described in the [last comment](https://github.com/ros2/rosidl/pull/183#issuecomment-307163340):

> While that might not need to be a use case which works in all possible combinations the behavior should at least be reasonable for the user (e.g. fail with an error message if the combination can't work).

esteve:
> I think the discussion in ros2/rosidl#183 summarizes already what the approach should be. The only thing missing is the behavior described in the last comment:

IMHO your comment describes what the end result should be, which I agree with, but now how it should be implemented. I proposed a concrete implementation (using `ament_ros_cmake` for messages), do you think that's the correct approach or that it should be done in a different way?

dirk-thomas:
> IMHO your comment describes what the end result should be, which I agree with, but not how it should be implemented. I proposed a concrete implementation (using `ament_ros_cmake` for messages), do you think that's the correct approach or that it should be done in a different way?

Using `ament_cmake_ros` for message packages is ok but I don't think it is sufficient atm in terms of detecting mixed mode builds. If that can be addressed by extending `ament_cmake_ros` that sounds good to me. I don't mind where this extra logic happens.

arm_crosscompilation/README.md
lmayencourt: no, after internal discussion we can remove this

lmayencourt:
> The last part that was missing for being able to build ROS2 statically (at least up to rcl)

@esteve as you say the current implementation allows to build up to `rcl`, but we still need modifications on `rclcpp`, `rcutils` and `examples` to build an example application with static libraries. 
I didn't consider the case of multiple typesupport packages and simply use `ament_cmake_ros` in the  `rosidl_xxx` packages but it is apparently not the good way. We have to find a solution for that...
In the meantime I can still prepare the PR for `rclcpp`, `rcutils` and `examples`.

esteve:
> @lmayencourt @esteve as you say the current implementation allows to build up to rcl, but we still need modifications on rclcpp, rcutils and examples to build an example application with static libraries.

I agree, I think it's best to at least being able to build ROS2 statically somehow, even if it's not the 100% perfect solution.

> In the meantime I can still prepare the PR for rclcpp, rcutils and examples.

I had patches lying around for those packages, I just submitted PRs for them, let me know if they are of any help to you:

https://github.com/ros2/common_interfaces/pull/43
https://github.com/ros2/rclcpp/pull/390
https://github.com/ros2/rcutils/pull/62
https://github.com/ros2/rosidl_typesupport/pull/14

esteve:
> @dirk-thomas Using ament_cmake_ros for message packages is ok but I don't think it is sufficient atm in terms of detecting mixed mode builds. If that can be addressed by extending ament_cmake_ros that sounds good to me. I don't mind where this extra logic happens.

Those are two separate issues IMHO, which are somewhat related, one is having a mechanism for messages to be statically built and the other handling mixed setups. I think the latter is a bit of a corner case, I imagine the users who want to build ROS2 statically, would do it all at once, and that'd also be a advanced use case that most users of ROS2 won't care much. Anyway, thanks for clarifying your comment.

mikaelarguedas:
I'm going to merge this given that it's ready to go (I'll update the README to point to the master branch though given that this branch will be deleted after merging). Feel free to comment here or on #306 to continue the discussion about static linking

mikaelarguedas:
@lmayencourt Can you confirm the main motivation for using the linaro cross toolchain rather than the gnu one ? Is it only to support non-debian based platforms or are there other specific advantages ?

lmayencourt:
We are using the linaro one internally, but I see no specific advantages compare to the gnu one.
@sudpau01  @davwan01 can probably give a better answer...

lmayencourt:
So apparently the linaro repository is just a collection of prebuilt GCC GNU binaries that are tested by Linaro . If you take the appropriate SHA from GNU and build the toolchain yourself, you should obtain exactly the same toolchain. 
I hope this answer help more !

PR #417, opened 2017-10-26
compile in a workspace rather than within the ros2 repository clone

mikaelarguedas:


mikaelarguedas:
> does the export CROSS_COMPILE= line that happens next need to change?

No it doesnt :confounded: https://github.com/ros2/ros2/pull/417/commits/e179fd4243d6fee414f55da205c23fa446f3a39c

lmayencourt:
@mikaelarguedas If you want to change the build directory, you also have to modify the path to the `install` directory in the `aarch64_toolchainfile.cmake`

arm_crosscompilation/README.md
```diff
@@ -16,9 +16,10 @@ Create a workspace and clone all repos:
 
 ```
 mkdir -p ros2_ws/src
-git clone https://github.com/ros2/ros2.git
-vcs-import ros2_ws/src < ./ros2/ros2.repos
 cd ros2_ws
+wget https://raw.githubusercontent.com/ros2/ros2/release-latest/ros2.repos
```
mikaelarguedas: Is targeting `release-latest` and not `master` anymore an intentional change ? (All our instructions point to the latest release so I think it's better like this, just want to make sure)

lmayencourt: yes, and I think we will have to change from `master` to `release-latest` the `wget` for the toolchain file when possible.

mikaelarguedas: :+1: sounds good to me, we should do that when we tag the next release. Thanks for the review

Issue #454, opened 2018-01-30
Ardent patch release 1


dirk-thomas:
For an Ardent patch release 1 we will cherry-pick the minimal changes which are necessary to address important issues. The high level goals are:
  * provide `rviz` for Windows
  * fix problem(s) in FastRTPS, e.g. https://github.com/eProsima/Fast-RTPS/issues/181
  * fix problem with include path order in `rosidl`
  * Provide `ros_environment` for the env vars specified in REP 149

We will do the following:

- [x] Create PRs using a to be created source branch named `ardent-patch1` targeting the existing `ardent branch`:
  - [x] `rviz` @wjwwood (https://github.com/ros2/rviz/pull/188)
  - [x] `pluginlib` @wjwwood (https://github.com/ros/pluginlib/pull/98)
  - [x] `resource_retriever` @wjwwood (https://github.com/ros/resource_retriever/pull/18)
  - [x] `fastrtps` @dirk-thomas (ros2/ros2#455)
  - [x] `rosidl` @mikaelarguedas (https://github.com/ros2/rosidl/pull/262)
  - `ros_environment` @dirk-thomas (this doesn't need a PR, the `ardent` branch is already the state we want to release)
- [x] Check if any packages should be updated on the macOS / Windows nodes (and if it affects the instructions): @wjwwood
- [x] Roll some fat archives to check they "address" the problems.
- [x] Merge the PRs.
- [x] Make a patch release in the affected repos.
- [x] Do the `bloom` releases.
  * @nuclearsandwich mentioned addressing a minor issue in the FastRTPS manifest which can be fixed along the way.
- [x] Upload the archives under a new release in the `ros2` repo.
- [x] Announce the patch release with a name like `ardent-2018-02-xx`.

wjwwood:
I created a single Windows CI job to see if `rviz` even compiles in this setup:

- Windows: [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_windows&build=4018)](http://ci.ros2.org/job/ci_windows/4018/)

nuclearsandwich:
> addressing a minor issue in the FastRTPS manifest which can be fixed along the way.

Addressed by https://github.com/ros2-gbp/fastrtps-release/pull/1

Noting for my own records that Fast-CDR hasn't changed since the ardent release so there's nothing there that needs bumping.

wjwwood:
[...]

PR #455, opened 2018-01-31
Ardent patch1

dirk-thomas:
The patch does the following for the upcoming patch release:

* Add `ros_environment` as a new package
* Use the latest hash of FastRTPS
  * While the [diff](https://github.com/eProsima/Fast-RTPS/compare/1f135d478dd9dcde98ed96c9aeef4f229e3a1450...bfb448e19bff253293e8f28ee8c3d3d78d882be2) isn't short it is the easiest way for us to get the fixes we need. All our tests are passing with `master` at the moment so I don't expect regressions. Therefore I would treat this as a "we sync with upstream of a third-party package" and expect they have done checking to avoid regressions.

Connect to ros2/ros2#454.

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -50,11 +50,11 @@ repositories:
   ros/pluginlib:
     type: git
     url: https://github.com/ros/pluginlib.git
-    version: 2.0.0
+    version: 3.0.0
   ros/resource_retriever:
     type: git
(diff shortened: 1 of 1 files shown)


dirk-thomas:
LGTM (I just can't review my own PR).

PR #463, opened 2018-02-20
advise to ask questions on ROS answers

mikaelarguedas:
This is the change I'm planning to make on all the issue templates. Opening only a single PR for feedback and once approved I'll push the same change to master on all repos that have one

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,5 +1,6 @@
 <!--
-For general questions, please post on discourse: https://discourse.ros.org/c/ng-ros
+For general questions, please ask on ROS answers: https://answers.ros.org
+For general design discussions, please post on discourse: https://discourse.ros.org/c/ng-ros
 For Bug report or feature requests, please fill out the relevant category below
 -->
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,5 +1,6 @@
 <!--
-For general questions, please post on discourse: https://discourse.ros.org/c/ng-ros
+For general questions, please ask on ROS answers: https://answers.ros.org
```
dirk-thomas: Should this mention the `ros2` tag?

mikaelarguedas: Yeah it should, I wasn't sure how to phrase it to let people know that they should use the tag corresponding to the rosdistro they use, and to use the `ros2` tag if they're unsure (as we didnt populate the `ROS_DISTRO` env var until the last patch release).

Any suggestion on phrasing is welcome

wjwwood: "For general questions, please ask on ROS answers and make sure to include at least the ros2 tag"?

tfoote: "For general questions, please ask on ROS answers and make sure to include at least the `ros2` tag and the rosdistro version you are running, aka `ardent`."

If there's only one release so far it's not too hard to guess the rosdistro version. (They could be running r2b3 too) but as we go forward it will be good to have these tagged so they can be cleaned up later, once the tagging is more ambiguous due to the subsequent ROS2 releases.

PR #477, opened 2018-03-31
add ros/angles and ros-perception/laser_geometry

wjwwood:
Connects to ros2/rviz#238

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -39,6 +39,10 @@ repositories:
     type: git
     url: https://github.com/eProsima/Fast-RTPS.git
     version: master
+  ros/angles:
+    type: git
+    url: https://github.com/ros/angles.git
+    version: ros2
   ros/class_loader:
     type: git
     url: https://github.com/ros/class_loader.git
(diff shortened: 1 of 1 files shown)


dhood:
merging to get builds working again

wjwwood:
Damn sorry, that’s the second time I’ve done that :/ thanks @dhood

dhood:
no problem, easy fixed

mikaelarguedas:
@wjwwood It looks like the angles package is not required anymore https://github.com/ros-perception/laser_geometry/commit/c029efcdd633df8b4935897cabeb95e31abddc14#diff-34ae68f4adad56c25c5bc05dcb64794e. Can we remove it from here ?
Should it be released anyway?

wjwwood:
If it’s not required I guess we can remove it.

mikaelarguedas:
:+1: 
I'll remove it from CI but release it in bouncy so that it's availab le to users and is tested on dev jobs. It just won't be tested on the other platforms anymore but will save some ci time

PR #486, opened 2018-04-23
remove ament_tools

dirk-thomas:
With the switch to `colcon` in ros2/ci#132 we don't need these two repositories anymore in the workspace.

[Build Status](https://ci.ros2.org/job/ci_linux/4285/)

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -15,10 +15,6 @@ repositories:
     type: git
     url: https://github.com/ament/ament_package.git
     version: master
-  ament/ament_tools:
-    type: git
-    url: https://github.com/ament/ament_tools.git
-    version: master
   ament/googletest:
     type: git
     url: https://github.com/ament/googletest.git


mikaelarguedas:
the CI code uses `osrf_pycommon` : https://github.com/ros2/ci/blob/e571eac5051c42a286885f8aca04200d5e20d1d7/ros2_batch_job/__main__.py#L28-L32

Removing it from our workspace reduces test surface for that package.

Are you planning on providing MacOS and Windows testing for this package on the upstream repository?
If not, maybe it's better to keep it in the workspace so that it keeps being tested as our infrastructure relies on it?

dirk-thomas:
> the CI code uses osrf_pycommon : https://github.com/ros2/ci/blob/e571eac5051c42a286885f8aca04200d5e20d1d7/ros2_batch_job/__main__.py#L28-L32

Thanks for clarifying.

> Are you planning on providing MacOS and Windows testing for this package on the upstream repository?

Since we only use the package in our CI infrastructure but don't expose it to ROS 2 users I don't see a reason why the ROS 2 jobs would need to test/cover this package.

Why is it the responsibility of ROS 2 to perform the testing of `osrf_pycommon` on these platforms? I would say that should happen with Travis / AppVeyor within that repo.

mikaelarguedas:
> Why is it the responsibility of ROS 2 to perform the testing of osrf_pycommon on these platforms? I would say that should happen with Travis / AppVeyor within that repo.

Agreed, hence the question:
> Are you planning on providing MacOS and Windows testing for this package on the upstream repository?

dirk-thomas:
> Are you planning on providing MacOS and Windows testing for this package on the upstream repository?

No, I am not going to do so. From my point of view this is not at all in the scope of ROS 2 to provide that kind of coverage for `osrf_pycommon`.

dirk-thomas:
Two additional concrete reasons why in my opinion we should not continue to test `osrf_pycommon`:

* If the code we are using for the CI scripts would not work on any of the platforms the builds would fail (or misbehave) before they even get to the point of running any tests on that package. So this doesn't achieve what you are describing.
* If the any of the code of `osrf_pycommon` which we *don't* use in our CI scripts doesn't work it doesn't matter for ROS 2 at all. So there is no reason why we should spend time on these tests in the first place.

mikaelarguedas:
We currently use a vendored version of `osrf_pycommon`, the day we upgrade we may face a lot of breakage if the package is not tested.
I think it's being good open-source citizen that to contribute to the projects we rely on. In this particular case we are maintaining that other package as well so removing already present test coverage means more work for us down the road without much benefit.

About the approach in general: if the build tool crashes on some packages for an unknown reason, removing packages that expose bugs is not a good path forward. The bug should be identified and fixed rather than ignored.

dirk-thomas:
> About the approach in general: if the build tool crashes on some packages for an unknown reason, removing packages that expose bugs is not a good path forward. The bug should be identified and fixed rather than ignored.

That is not what is happening here. This PR removes a package which we simply don't need and use anymore. Nothing more - nothing less. The fact that their was a failing build with that specific package has nothing to do with that rational.

And the problem is not ignored at all. The PR colcon/colcon-core#25 is an active step to debug that specific problem. The build failed once and it is unclear why that is the case. The information added in that patch will hopefully help *if* the problem happens again.

> Currently all our instructions rely on ament_tools being part of the workspace, merging this without updating the documentation would break any user using master.

Sounds good to me. Waiting for the wiki pages to be updated then before merging this (some time) after that.

dirk-thomas:
Putting this back into "ready" since we will only merge this a archive `ament_tools` after the Bouncy release.

dhood:
heads up so it doesn't go unnoticed: https://github.com/ros2/launch/pull/74 reintroduces a osrf_pycommon dependency

wjwwood:
Thanks for the reminder @dhood. This feel off my radar when I saw it was bumped until after the release. At that point we can consider removing it again or moving it or something.

dirk-thomas:
With Bouncy released `ament_tools` can be removed from the repos file now since it won't be released into future releases.

`osrf_pycommon` is not being touched anymore (despite the branch name).

wjwwood:
I'll make a follow up to relocate `osrf_pycommon` to the `osrf` folder since it's not used by anything in the `ament` folder anymore.

dvogureckiy99:
And how know compile catkin packages ? You even have used example package that can be build only with catkin https://docs.ros.org/en/humble/Tutorials/Intermediate/URDF/Building-a-Visual-Robot-Model-with-URDF-from-Scratch.html ?

PR #500, opened 2018-06-06
add navigation_msgs repository (for map_msgs package)

wjwwood:
`move_base_msgs` is disabled for now, since it only has Actions in it.

Connects to ros2/rviz#267


dirk-thomas:
Please see the follow up issue ros-planning/navigation_msgs#6.

PR #524, opened 2018-07-05
use console_bridge_vendor and remove console_bridge repository

mikaelarguedas:
connects to ros2/console_bridge_vendor#1


mikaelarguedas:
@ros2/team FYI: this will require removing `ros/console_bridge` from your workspace and to import the new repos file

PR #557, opened 2018-08-03
Add crystal milestone and descriptions to labels

sloretz:
This adds a crystal milestone and label descriptions. I copied the descriptions for: `bug`, `duplicate`, `enhancement`, `good first issue`, `help wanted`, `invalid`, `question`, `wontfix` from the defaults on a new-ish github repo. I made up the descriptions for `more-information-needed`, `ready`, `in progress`, `in review`.

@dirk-thomas the script defaults to `ament` and `ros2` orgs. Would you like it applied to `colcon` as well? I suspect not since colcon isn't tied to the `crystal` release.

PR #564, opened 2018-08-16
Crystal release December 14

sloretz:
Fix crystal release date https://github.com/ros2/ros2/pull/557#pullrequestreview-146290839


sloretz:
Log of update with this PR

[aug_16_2018_crystal_date.log](https://github.com/ros2/ros2/files/2295247/aug_16_2018_crystal_date.log)

It looks like the script always sees a label's current description as being `None`, so it always sets it. I ran the script anyways because it does no harm other than take longer to complete.

PR #565, opened 2018-08-22
update tags for bouncy patch 2

mikaelarguedas:


mikaelarguedas:
Always good to have another set of :eyes: :+1:

PR #622, opened 2018-12-13
insert osrf_pycommon inside osrf folder

abilbaotm:
actually osrf_pycommon is set inside the ament folder, but the correct organization is osrf.

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -19,10 +19,6 @@ repositories:
     type: git
     url: https://github.com/ament/googletest.git
     version: ros2
-  ament/osrf_pycommon:
-    type: git
-    url: https://github.com/osrf/osrf_pycommon.git
-    version: master
   ament/uncrustify_vendor:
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
> Makes sense to me, but I'm not sure if we'll want to hold this or not until after we finish doing the Crystal release. @nuclearsandwich

These directories are cosmetic / for human benefit rather than computer. But in the interest of not touching pieces that don't need to be touched let's wait on this.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5903/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2446/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4891/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5807/)

Issue #638, opened 2019-01-03
ROS 2 Crystal Clemmys Patch Release 1


clalancette:
This is a meta-issue to track the changes that are going into the first patch release for Crystal Clemmys.  @ros2/team , please add PRs that are going onto the Crystal branches to this ticket for tracking.  @nuclearsandwich FYI.

PRs merged for Crystal Patch Release 1:
-
-

dirk-thomas:
I would propose to separate PRs which actually fix bugs and should go into the patch release from the ones which we might sneak in since it just makes life easier - at least visually. Only the first group is basically the reason why we do patch releases. Especially in the `rmw_connext` and `rmw_fastrtps` repos almost all except one per repo seem to be of the later category.

Also the `rosout` related changes are introducing a new feature. In general we only aim to release those into future distros. Otherwise it would require extensive testing to ensure it does neither change or break any existing behavior. I am not saying we can't make an exception but just want to state that this is the general policy in all ROS distros and we should have a good reason to diverge from that.

nuclearsandwich:
> I would propose to separate PRs which actually fix bugs and should go into the patch release from the ones which we might sneak in since it just makes life easier

It seems that we want two different things from this list. It sounds like you want the list of PRs which justify the patch release which serve as a draft "release notes/highlights". Whereas I want a list of every change that would be released in the next sync on the coattails of the PRs from the short list so that they can be reviewed for potentially breaking changes. I'll see about hiding these "coattail riders" behind a `<details>` tag once I finish the complete list.

nuclearsandwich:
I split out the main body PR. I haven't yet made an explicit list of coattail riders, but it's everything in the last list that isn't mentioned in the first.

nuclearsandwich:
Releases still need to be made for
- ament/ament_lint
- ros2/demos
- ros2/sros2
- ros2/rviz

Other repositories have been released with their desired changes from the list of merged PRs summarized by feature/fix and the PRs which rode their coattails into this release. Except to fix major issues, repositories which have been tagged for this patch release will not be re-released until the next patch that has a desired change for that repository.

wjwwood:
Here's the ros2/rviz release: https://github.com/ros/rosdistro/pull/19943

nuclearsandwich:
Packages are in the testing repository for Ubuntu and #646 contains a repos file of the prospective patch release for people to try.

nuclearsandwich:
[...]

PR #646, opened 2019-01-16
Update Crystal repos file for Patch release 1.

nuclearsandwich:
This should be what we release on Thursday (17 Jan) if no regressions are found.

Connects to #638

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -10,7 +10,7 @@ repositories:
   ament/ament_lint:
     type: git
     url: https://github.com/ament/ament_lint.git
-    version: 0.6.2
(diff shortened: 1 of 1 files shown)


jacobperron:
Maybe it was overlooked, but there were a couple fixes added to `rmw_fastrtps` (at least one listed in the description #638) that are not present in the patch release. If the ship has sailed, I think it's okay if they wait for the next patch release/sync.

nuclearsandwich:
Addressed in https://github.com/ros2/ros2/issues/638#issuecomment-454625824 to keep discussion centralized on that issue.

PR #659, opened 2019-02-21
Pin FastRTPS pre eProsima/Fast-RTPS/pull/411

sloretz:
Upstream changes are causing test failures (ros2/build_cop#166). This pins it temporarily until the problem is resolved.


sloretz:
Still one more test failure to solve, but pinning the fast-rtps version isn't the right solution since it appears to be a test code issue.

PR #676, opened 2019-03-21
Add ros2/launch_ros repository to ros2.repos

hidmic:
Precisely what the title says.

This PR depends on ros2/launch_ros#1 and ros2/launch#211 and as such should only be merged after them.

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -151,6 +151,10 @@ repositories:
     type: git
     url: https://github.com/ros2/launch.git
     version: master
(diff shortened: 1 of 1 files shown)


hidmic:
I'll get the word out once this is merged.

wjwwood:
> This PR depends on ros2/launch_ros#1 and ros2/launch#211 and as such should only be merged after them.

It should be merged at the same time as them, otherwise people will be missing required packages from `ros2/launch`.

PR #687, opened 2019-04-10
Update with instructions to build OpenCV packages

mjcarroll:
These are the steps that I used to build 3.4.6 for VS2019.

Signed-off-by: Michael Carroll <michael@openrobotics.org>

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,15 @@
 
 See [the ROS 2 wiki](https://github.com/ros2/ros2/wiki/Windows-Development-Setup#install-a-binary-distribution-of-opencv) for installation instructions.
 
+
+## For ROS2 Crystal and current development with VS2019
+
+* Visual Studio 2019: https://github.com/ros2/ros2/releases/download/opencv-archives/opencv-3.4.6-vc16.VS2019.zip
+
+
 ## For ROS 2 Bouncy Bolson and current development
 
-Visual Studio 2017: https://github.com/ros2/ros2/releases/download/opencv-archives/opencv-3.4.1-vc15.VS2017.zip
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,9 +2,15 @@
 
 See [the ROS 2 wiki](https://github.com/ros2/ros2/wiki/Windows-Development-Setup#install-a-binary-distribution-of-opencv) for installation instructions.
 
+
+## For ROS2 Crystal and current development with VS2019
```
dirk-thomas: This is only for Dashing+.

The VS 2017 one is for Bouncy and Cystal.

mjcarroll: Fixed in latest commit.

README.md
```diff
@@ -2,9 +2,15 @@
...
 
-Visual Studio 2017: https://github.com/ros2/ros2/releases/download/opencv-archives/opencv-3.4.1-vc15.VS2017.zip
+## For ROS2 Dashing and current development
+
+* Visual Studio 2019: https://github.com/ros2/ros2/releases/download/opencv-archives/opencv-3.4.6-vc16.VS2019.zip
+* Visual Studio 2017: https://github.com/ros2/ros2/releases/download/opencv-archives/opencv-3.4.6-vc15.VS2017.zip
+
+## For ROS 2 Bouncy Bolson and Crystal Clemmys
+
+* Visual Studio 2017: https://github.com/ros2/ros2/releases/download/opencv-archives/opencv-3.4.1-vc15.VS2017.zip
```
dirk-thomas: We can only install one of the OpenCV version on our Windows nodes. So for future Crystal patch releases we will need to use the same one as for Dashing (if we don't want to keep separately configured Windows machines around).

PR #704, opened 2019-05-15
Add rostest repository to ros2.repos file

hidmic:
Exactly what the title says. Connected to https://github.com/ros2/launch/issues/208.


wjwwood:
I should probably rename this repository first right?

hidmic:
Same as in https://github.com/ros2/rostest/pull/1#issuecomment-493067685, no strong preference on my end.

PR #707, opened 2019-05-21
use branch names on the Crystal branch

dirk-thomas:
Instead of tags use the upstream development branch. The tagged version is available from the `release-crystal-<TS>` tags in this repo.

This will enable the nightly CI build for Crystal to actually show current state before making new releases.

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -2,31 +2,31 @@ repositories:
   ament/ament_cmake:
     type: git
     url: https://github.com/ament/ament_cmake.git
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
My only reservation with updating the semantic of the `crystal` branch from "Latest releases of Crystal" to "Latest development branches of Crystal" is that there may be some who are using this branch to get the latest release regardless of tag. Would separate "crystal" and "crystal-devel" branches work to give us both options?

PR #748, opened 2019-07-24
Add tracing tools repo

christophebedard:
Signed-off-by: Christophe Bedard <fixed-term.christophe.bourquebedard@de.bosch.com>
Signed-off-by: Christophe Bedard <bedard.christophe@gmail.com>
Signed-off-by: Christophe Bedard <christophe.bedard@apex.ai>

Related to https://github.com/ros2/rcl/pull/473
Related to https://github.com/ros2/rclcpp/pull/789

With instrumentation, the `tracetools` package becomes a core dependency.

cc @iluetkeb

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -35,6 +35,10 @@ repositories:
     type: git
     url: https://github.com/eProsima/foonathan_memory_vendor.git
     version: master
+  micro-ROS/ros_tracing/ros2_tracing:
+    type: git
+    url: https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing.git
+    version: 0.2.8
   osrf/osrf_pycommon:
     type: git
     url: https://github.com/osrf/osrf_pycommon.git


Karsten1987:
Are there any special flags which should be passed to cmake for the tracingtools? I've seen on the gitlab page some instructions of `-DWITH_LTTNG=ON`.

iluetkeb:
That flag was for the ROS 1 version of tracetools. In this version, tracing is automatically enabled *if* LTTng is found by CMake.

The way we see this work right now is that somebody who wants to trace just installs LTTng, and then only has to recompile tracetools in the local workspace -- that's it. In the future, we would like to discuss whether tracing could always be enabled (tracing is purposefully very little overhead), but that's a matter for latter.

ros2.repos
```diff
@@ -315,6 +315,10 @@ repositories:
     type: git
     url: https://github.com/ros2/tlsf.git
     version: master
+  ros2/tracing:
```
wjwwood: We like to keep the folder structure matching the github/gitlab organization layout, so this ought to be in `micro-ROS/ros2_tracing`.

wjwwood: Since you have sub-groups in your gitlab project, maybe `micro-ROS/ros_tracing/ros2_tracing` is more appropriate.

christophebedard: Fixed it!

iluetkeb:
Regarding breaking CI: We can use a tag or a branch instead of master. If we use a tag, that means a lot of PR's for you guys, but it's totally safe, since you know it works. If we use a branch, it's a bit more risk but not much. Thus, I would prefer using a branch. Okay?

Regarding control: What would you need to feel like you're in control? Should we put this into rcl directly (it's not big, and all dependencies are optional)? Should we have the repo in the ros2 org?

Regarding performance: None of the tracepoints are in critical places that are called very often. No inner loops or something the like. In principle, if you have a timer or the like that's extremely short, the tracepoint could add a measureable (though still not significant) overhead. In general, though, performance is a subject we could certainly look at more, if you're worried about it.

wjwwood:
> If we use a tag, that means a lot of PR's for you guys, but it's totally safe, since you know it works. If we use a branch, it's a bit more risk but not much. Thus, I would prefer using a branch. Okay?

The nice thing about a pr for each time a new version comes up is that it's an opportunity for us to run CI on the new state and compare it to the old. The goal is really to minimize downtime and "known flaky or failing tests" on our "master" so that other on going work is not disrupted.

I would personally prefer using tagged versions, especially since I don't expect this dependency to be changing frequently (at least not over the long term), but that's just my opinion. Maybe others on the team would feel differently. In contrast, we put up with the downsides of using master from Fast-RTPS because it still changing so much and we often need intermediate fixes (fixes between releases).

Alternatively, if every change to your repository had to pass our CI before being merged to master, that would be ok too. But that puts a continuous burden on you guys and taxes our CI resources as well.

> Regarding control: What would you need to feel like you're in control? Should we put this into rcl directly (it's not big, and all dependencies are optional)? Should we have the repo in the ros2 org?

This is really a non-issue for me. I'm totally fine for you guys to own it and host it. The point is that we may choose to fork it or pin the version to an older release when doing our own release if we run into issues. Whereas if it were one of our packages we would likely fix the problem and author a new release right then and there. But again this is only a problem if there is an issue and you guys cannot respond to the problems quickly, neither of which I anticipate over the long run.

My opinion is that nothing needs to be done here.

> Regarding performance: None of the tracepoints are in critical places that are called very often. No inner loops or something the like. In principle, if you have a timer or the like that's extremely short, the tracepoint could add a measureable (though still not significant) overhead. In general, though, performance is a subject we could certainly look at more, if you're worried about it.

I definitely don't want to do premature optimization, and I think having the call out to your library is fine in most cases, even like our official Debian packages, but at the same time I think it should be possible for someone building ROS 2 from source to completely compile out the trace lines from rcl and rclcpp (and even avoid needing to build or link your library) if they decide they want to do so.

iluetkeb:
Alright, we'll tag it and update this PR.

> I definitely don't want to do premature optimization, and I think having the call out to your library is fine in most cases, even like our official Debian packages, but at the same time I think it should be possible for someone building ROS 2 from source to completely compile out the trace lines from rcl and rclcpp (and even avoid needing to build or link your library) if they decide they want to do so.

Disabling this completely would be relatively trivial, since "TRACEPOINT" is already a macro, so we could add a compile time setting to the tracetools and define it as empty. 

Alternatively, we could conditionally include "tracetools.h" in rcl/rclcpp, depending on a compile-time option in those packages. This would remove the need to even have tracetools in the workspace, but at the cost of littering rcl/rclcpp includes with ugly ifdefs.

Your choice, though I would strongly favor the former.

wjwwood:
Yeah, making it completely optional is probably not necessary (at least until someone asks for it), but I do want to make sure that people compiling from source can completely compile it out, which it sounds like we could do in the header of `ros2_tracing`.

nuclearsandwich:
> Alright, we'll tag it and update this PR.

Is this still happening? If so I'll wait to add my :heavy_check_mark: until then. And it would make sense to test the new package passes CI.

wjwwood:
> > Alright, we'll tag it and update this PR.
> Is this still happening?

Oh yeah, that would be good to do first. Obviously passing CI too.

wjwwood:
Not sure if we need to merge this with the above mentioned pr's at the same time or not, but here's CI just up to the `tracetools*`/`ros2trace` packages to see how it looks so far:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7726/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3830/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6328/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7585/)

christophebedard:
The failure on Windows should be fixed by [ros2_tracing!71](https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing/merge_requests/71) ([decoding issue](https://ci.ros2.org/job/ci_windows/7585/consoleFull#console-section-91) with "ü"/`0xfc` in `setup.py` files). Updated to `0.2.1` :smile:

iluetkeb:
btw... I think you may have a problem on your Windows build farm. The "setup.py" is utf-8 and contained an utf-8 special character, which works fine on Linux, but not on Windows. Usually this indicative of some terminal not having a utf-8 encoding when you're expecting it to be.

Christophe found an interesting [stackoverflow entry related to terminal encoding](https://stackoverflow.com/a/42389786/6476709), which also specifies how you can use the terminal's encoding on stdin.

dirk-thomas:
> I think you may have a problem on your Windows build farm.

I don't think the problem is on the buildfarm. It is reasonable to expect the build to pass independently what locale a machine has set. The logic extracting the metadata in from the `setup.py` file should work independent of the default encoding. colcon/colcon-python-setup-py#19 tries to achieve that:

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7607/)
  * While the build still fails it resolves the encoding problem.
  * The remaining failure might need to be addressed in `tracetools`.

Therefore I would suggest to revert the changes to replace the non-ASCII characters in your repo.

iluetkeb:
> The remaining failure might need to be addressed in tracetools.

Looks like it. However, I'm a bit puzzled: The library is correctly created as a dynamic library ("tracetools.dll") but then the build for the status tool tries to read "tracetools.lib", which would be the name if the library were static.

Do you have an idea why that happens?

We unfortunately do not have a Windows build machine available locally right now, so it's a bit hard to test for us.

wjwwood:
Yeah, as you said, usually this is due to lacking dll export/import attributes on your classes and functions.

wjwwood:
New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7774/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3864/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6353/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7616/)

iluetkeb:
The unicode decode error is back...

We reverted our change on @dirk-thomas 's request, but apparently the fix in colcon isn't working?

christophebedard:
I think it's not using the right `.repos` file. It's testing b9ff095/`0.2.0` (separate from the encoding issue).

iluetkeb:
> I think it's not using the right .repos file. It's testing b9ff095/0.2.0 (separate from the encoding issue).

Where did you see that @christophebedard ?

christophebedard:
> > I think it's not using the right .repos file. It's testing [b9ff095](https://github.com/ros2/ros2/commit/b9ff095e2bef4cadfa5cec03b7ef49d56c530e7d)/0.2.0 (separate from the encoding issue).
> 
> Where did you see that @christophebedard ?

The job's build information:

> repos_url: https://raw.githubusercontent.com/ros2/ros2/b9ff095e2bef4cadfa5cec03b7ef49d56c530e7d/ros2.repos

dirk-thomas:
> I think it's not using the right .repos file.

Correct, the passed hash is still pointing to the 0.2.0 release. I will trigger another build using the latest: https://raw.githubusercontent.com/ros2/ros2/f05f0d2183608d491d9fa709dd35589871e13657/ros2.repos

> but apparently the fix in colcon isn't working?

The fix has only been merged into `master` but not yet released in a new package version. Therefore the job needs to be triggered with explicitly using `master` branches from `colcon` (rather than the released packages):

* Windows [Build Status](https://ci.ros2.org/job/ci_windows/7617/)

wjwwood:
Sorry, that's my bad, copied the repo's file from the old test without thinking about it.

wjwwood:
Looks like the Windows job has some compiler warnings still, but otherwise it's looking ok.

christophebedard:
> Looks like the Windows job has some compiler warnings still, but otherwise it's looking ok.

Fixed with `0.2.3`!

dirk-thomas:
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/7628/)

christophebedard:
Still warnings :/

iluetkeb:
Those are just “unused parameter”, because the functions are empty. We turned those off when compiling without LTTng on gcc, but MSVC would likely require different pragmas.

This is harmless and can be ignored IMHO.

dirk-thomas:
> This is harmless and can be ignored IMHO.

Those compiler warnings result in the job being unstable. We can't merge any changes which don't have all jobs being stable. So while they might be "harmless" they still need to be resolved before this can be considered for merging.

iluetkeb:
Understood. It’s end of day here, so we’ll fix that tomorrow.

wjwwood:
Yeah, it helps us detect regressions, so we require it to be completely green, even if it's just disabling specific warnings (please disable the specific warning in the narrowest scope so we don't produce false positives in the future).

Thanks!

christophebedard:
Sorry for the delay.

Updated to `0.2.4`. There shouldn't be any warnings now :tm:

dirk-thomas:
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/7753/)

christophebedard:
> There shouldn't be any warnings now :tm:

Well technically there were no warnings :laughing:

Updated to `0.2.5`.

dirk-thomas:
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/7762/)

christophebedard:
Updated to `0.2.6`.

dirk-thomas:
The Windows build is looking good: [Build Status](https://ci.ros2.org/job/ci_windows/7771/)

Running a full set:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7921/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3955/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6448/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7772/)

iluetkeb:
I'm not sure the "ament_xmllint not found" is our fault... Shouldn't this always be there?

Also, how can it be that first that first the Windows passes, and on the next try it's unstable?

In any case, Christophe has finished his internship now and I'm on vacation for two weeks, so both of us won't be looking at this until at least then... See you then!

dirk-thomas:
> I'm not sure the "ament_xmllint not found" is our fault... Shouldn't this always be there?

You packages uses it in a test but doesn't declare that test dependency in the manifest.

> Also, how can it be that at first the Windows build passes, and on the next try it's unstable?

The Windows build before was only testing your specific packages. The last build tested all ROS 2 packages. Those failing tests aren't related to your code and you can ignore them - they are failing on master already.

christophebedard:
> Your package uses it in a test but doesn't declare that test dependency in the manifest.

Oops my bad!

dirk-thomas:
@christophebedard Friendly ping. Is there an updated version we should test again?

christophebedard:
@dirk-thomas sure, could you test this one pointing to a specific commit with a fix? https://gitlab.com/micro-ROS/ros_tracing/ros2/raw/a4a32c7b653014a69686713a4b191550669f25b4/ros2.repos

I think @iluetkeb just got back from vacation, so we may be able to create a new release sometime soon.

dirk-thomas:
> could you test this one pointing to a specific commit with a fix?

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8083/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4088/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6588/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7962/)

iluetkeb:
So, this is looking good, hmm?

I've reviewed Christophe's merge requests, all okay there as well.

christophebedard:
Updated to `0.2.7`.

dirk-thomas:
> Updated to `0.2.7`.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8099/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4098/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6600/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7981/)

iluetkeb:
At first glance the Windows test failure do not look related to our work. Or am I missing something?

dirk-thomas:
> At first glance the Windows test failure do not look related to our work. Or am I missing something?

@cottsay I will let Scott comment if all those test failures are expected.

cottsay:
> I will let Scott comment if all those test failures are expected.

Looks good to me.

christophebedard:
@wjwwood @nuclearsandwich @cottsay @dirk-thomas I've updated the version to `0.2.8` following the changes made for https://github.com/ros2/rclcpp/pull/789 and I also squashed everything.

wjwwood:
> I've updated the version to `0.2.8`

Here's new CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8388/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4305/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6830/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8286/)

wjwwood:
I think this pr needs to be rebases, since we had to update the repos file already to pin the version of fastrtps...

christophebedard:
> I think this pr needs to be rebases, since we had to update the repos file already to pin the version of fastrtps...

Done!

wjwwood:
Ok, CI again:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8391/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4307/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6832/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8291/)

christophebedard:
Seems like test failures are not related to this :smiley:

wjwwood:
@cottsay can you check the latest CI and confirm those are expected failures?

cottsay:
> @cottsay can you check the latest CI and confirm those are expected failures?

I don't know if I'd use the term "expected", but we're seeing all of those in nightlies.

wjwwood:
Ok, I guess this is good to merge then.

wjwwood:
I think https://github.com/ros2/rcl/pull/473 and https://github.com/ros2/rclcpp/pull/789 are not needed to merge this pr, if anyone thinks that's not right, let me know.

PR #780, opened 2019-09-13
[dashing] add turtlesim

dirk-thomas:
Same as #778.


dirk-thomas:
There shouldn't be any side effects. Worst case the package is included in the fat archives but not available from Debian packages. Before doing a release it needs updated artwork...

PR #781, opened 2019-09-19
add new python_cmake_module repository

wjwwood:
This is needed after we migrate the `python_cmake_module` out to its own repository.

Connects to ros2/rosidl_python#87

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -195,6 +195,10 @@ repositories:
     type: git
     url: https://github.com/ros2/poco_vendor.git
     version: master
+  ros2/python_cmake_module:
+    type: git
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Are we planning to backport this split to Dashing to make future backports easier or do we not want to change the repository layout in Dashing but make future backports more effort?

nuclearsandwich:
There is relatively low activity on python_cmake_module (there are almost as many commits bumping the version as making actual content changes) so I think the chance that this repository will see backported changes is low enough to justify not backporting the repository split.

hidmic:
@wjwwood is there anything blocking this PR and https://github.com/ros2/rosidl_python/pull/87? Or can we merge it?

PR #797, opened 2019-10-08
Merge robot_state_publisher back to original repository.

clalancette:

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -257,7 +257,7 @@ repositories:
     version: master
   ros2/robot_state_publisher:
     type: git
-    url: https://github.com/ros2/robot_state_publisher.git
+    url: https://github.com/ros/robot_state_publisher.git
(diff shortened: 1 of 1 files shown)


ros2.repos
```diff
@@ -257,7 +257,7 @@ repositories:
     version: master
   ros2/robot_state_publisher:
```
dirk-thomas: Should the target directory be updated too?

clalancette: Yeah, it wasn't totally clear to me if we should move it back to the `ros` directory or not.  I don't have a strong opinion either way.

dirk-thomas: If there is no strong reason not to I would suggest to move it to the matching directory location.

Karsten1987: addressed in here: https://github.com/ros2/ros2/pull/803

PR #800, opened 2019-10-15
Pin Fast-RTPS

ivanpauno:
After https://github.com/eProsima/Fast-RTPS/pull/775, it seems that builds are failing.

e.g.:
https://ci.ros2.org/job/ci_linux/8359/console#console-section-11
https://ci.ros2.org/job/ci_linux-aarch64/4290/console#console-section-9
https://ci.ros2.org/view/nightly/job/nightly_linux-armhf_repeated/259/consoleFull#console-section-264

Pinning until the problem is solved upstream.

PR #801, opened 2019-10-16
Revert "Pin Fast-RTPS (#800)"

ivanpauno:
Reverts #800.

Upstream ticket was solved https://github.com/eProsima/Fast-RTPS/issues/785.

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -30,7 +30,7 @@ repositories:
   eProsima/Fast-RTPS:
     type: git
     url: https://github.com/eProsima/Fast-RTPS.git
-    version: 63359c20057818feb0476b316d668ce8c898c31a
+    version: master
   eProsima/foonathan_memory_vendor:
     type: git
     url: https://github.com/eProsima/foonathan_memory_vendor.git


ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8385/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4303/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6829/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8284/)

dirk-thomas:
Should we lock Fast-RTPS to the latest release from yesterday 1.9.2 to decouple us from their master changes?

ivanpauno:
> Should we lock Fast-RTPS to the latest release from yesterday 1.9.2 to decouple us from their master changes?

The only problem is that that release doesn't include https://github.com/eProsima/Fast-RTPS/pull/786, and our builds are failing without that.
https://github.com/eProsima/Fast-RTPS/pull/779 introduced that build error, and it's included in 1.9.2 (which is also good, because it fixes https://github.com/ros2/build_cop/issues/246).

Maybe pinning to https://github.com/eProsima/Fast-RTPS/commit/f95cb0228487acb9aa8eb77ddfc85de51215933d or https://github.com/eProsima/Fast-RTPS/commit/3755c98e23e7137b09167c39f1cc9195054ddcbf.
I think it's a good idea to pin Fast-RTPS until we release Eloquent, and we have a separated branch for it.

dirk-thomas:
@richiprosima @JaimeMartin Are you planning to tag a working version 1.9.3 very soon (with 1.9.2 broken)?

hidmic:
I'd say, let's get this in and pin. `ros2/system_tests` failures are gone with this patch.

cottsay:
FWIW, the last commit before we started getting all of these test failures and build breaks was eProsima/Fast-RTPS@9d56202 as part of this build: [Build Status](https://ci.ros2.org/view/nightly/job/nightly_linux_debug/1337/)

My suggestion would be to pin there.

dirk-thomas:
Does `master` work fine too?

hidmic:
> Does `master` work fine too?

As of today, yes it does.

cottsay:
Ah, I misunderstood - ignore my earlier suggestion.

+1 for pinning to eProsima/Fast-RTPS@f95cb02 or eProsima/Fast-RTPS@3755c98

ivanpauno:
Merging the revert. Why can decide where to pin Fast-RTPS in a follow-up PR, atm `master` is working well.

JaimeMartin:
Hi guys, we are looking at this.

+ @raquelalvarezbanos @MiguelCompany

MiguelCompany:
> > Should we lock Fast-RTPS to the latest release from yesterday 1.9.2 to decouple us from their master changes?
> 
> The only problem is that that release doesn't include [eProsima/Fast-RTPS#786](https://github.com/eProsima/Fast-RTPS/pull/786), and our builds are failing without that.
> [eProsima/Fast-RTPS#779](https://github.com/eProsima/Fast-RTPS/pull/779) introduced that build error, and it's included in 1.9.2 (which is also good, because it fixes [ros2/build_cop#246](https://github.com/ros2/build_cop/issues/246)).
> 

Just for the record, eProsima/Fast-RTPS#779 **did not** introduce the build error, it was eProsima/Fast-RTPS#775. So the build error was introduced in master (and fixed later by eProsima/Fast-RTPS#786), not in 1.9.x

Master is fixed now, and we have instructed new members in the team to build and test ROS2 when reviewing PRs similar to 775.

Nevertheless, as master is already in the road to v2.0, where DDS API will be introduced, I would suggest to pin Fast-RTPS to branch 1.9.x, where we don't expect big changes to occur, except backporting of some bug-fixes and allocation reducing features. We have scheduled [Release 1.9.3](https://github.com/eProsima/Fast-RTPS/issues/800) for Nov 7th.

ivanpauno:
> Just for the record, eProsima/Fast-RTPS#779 did not introduce the build error, it was eProsima/Fast-RTPS#775. So the build error was introduced in master (and fixed later by eProsima/Fast-RTPS#786), not in 1.9.x

Sorry, my bad. I confused the two.

> Nevertheless, as master is already in the road to v2.0, where DDS API will be introduced, I would suggest to pin Fast-RTPS to branch 1.9.x

Sounds like a good idea to me.

PR #808, opened 2019-10-21
move CycloneDDS repos into ros2.repos file

dirk-thomas:
WIth CycloneDDS having the same Tier level as OpenSplice and ros2/ci#340 there is no need for a separate `.repos` file anymore.


dirk-thomas:
While it partially overlaps with the merge PR #787 it can still be merged to delete the obsolete separate CycloneDDS.repos file.

PR #810, opened 2019-10-21
Remove rclc and ros2/tutorials.

clalancette:
They've been disabled for literally years now; I think it is
safe to say we aren't actively working on them.  If we want
to resurrect them in the future, we can always add them back.


nuclearsandwich:
comments in YAML make automating generation much more difficult to do.

PR #832, opened 2019-12-02
point ros_tutorials to eloquent-devel branch

mikaelarguedas:
Signed-off-by: Mikael Arguedas <mikael.arguedas@gmail.com>


dirk-thomas:
Thanks.

PR #833, opened 2019-12-02
add rqt_action and rqt_topic

mikaelarguedas:
Signed-off-by: Mikael Arguedas <mikael.arguedas@gmail.com>

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -23,6 +23,10 @@ repositories:
     type: git
     url: https://github.com/ament/uncrustify_vendor.git
     version: master
+  eclipse-cyclonedds/cyclonedds:
+    type: git
+    url: https://github.com/eclipse-cyclonedds/cyclonedds.git
+    version: master
   eProsima/Fast-CDR:
     type: git
     url: https://github.com/eProsima/Fast-CDR.git
@@ -34,10 +38,6 @@ repositories:
   eProsima/foonathan_memory_vendor:
     type: git
     url: https://github.com/eProsima/foonathan_memory_vendor.git
-    version: master
-  eclipse-cyclonedds/cyclonedds:
-    type: git
-    url: https://github.com/eclipse-cyclonedds/cyclonedds.git
     version: master
   micro-ROS/ros_tracing/ros2_tracing:
     type: git
--- a/ros2.repos
+++ b/ros2.repos
@@ -51,17 +51,9 @@ repositories:
     type: git
     url: https://github.com/osrf/osrf_testing_tools_cpp.git
     version: master
-  ros-perception/image_common:
-    type: git
-    url: https://github.com/ros-perception/image_common.git
-    version: ros2
   ros-perception/laser_geometry:
     type: git
     url: https://github.com/ros-perception/laser_geometry.git
-    version: ros2
-  ros-perception/vision_opencv:
-    type: git
-    url: https://github.com/ros-perception/vision_opencv.git
     version: ros2
   ros-planning/navigation_msgs:
     type: git
@@ -94,10 +86,6 @@ repositories:
   ros-visualization/rqt_graph:
     type: git
     url: https://github.com/ros-visualization/rqt_graph.git
-    version: crystal-devel
-  ros-visualization/rqt_image_view:
-    type: git
-    url: https://github.com/ros-visualization/rqt_image_view.git
     version: crystal-devel
   ros-visualization/rqt_msg:
     type: git
--- a/ros2.repos
+++ b/ros2.repos
@@ -75,6 +75,10 @@ repositories:
     type: git
     url: https://github.com/ros-visualization/rqt.git
     version: crystal-devel
+  ros-visualization/rqt_action:
+    type: git
+    url: https://github.com/ros-visualization/rqt_action.git
+    version: crystal-devel
   ros-visualization/rqt_console:
     type: git
     url: https://github.com/ros-visualization/rqt_console.git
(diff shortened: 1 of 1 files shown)


dirk-thomas:
What is the benefit of adding this package to the repos file?

mikaelarguedas:
This ensures all the packages listed in the metapackage (that will be part of the variants) are automatically tested on all platforms

dirk-thomas:
Since this is an empty "meta" package I don't think it is necessary to add it here. For Debian packages this mostly exists for ease of installation which isn't applicable for the big archives built from the repos file.

jacobperron:
I think this will cause problems with some of our CI jobs  since not all of the packages listed in the metapackage are in the repos file (e.g. `rqt_web`). `rosdep` will fail for example.

mikaelarguedas:
My point was that this ensures that all packages listed in this `package.xml` are also part of the repos file, so ensuring these packages are part of the various CI. For example this will ensure that a run of rosdep would fail if the packages failed to be added to this repos file.

Then as the CMakeLists of the "meta" package is empty and there is no run of rosdep or other mechanism to ensure dependencies' presence I guess you're right it will not be useful as it won't detect untested dependencies

dirk-thomas:
> I think this will cause problems with some of our CI jobs  since not all of the packages listed in the metapackage are in the repos file (e.g. `rqt_web`).

Specifically `rqt_web` has not been ported and is therefore also not listed as dependency in `rqt_common_plugins` atm: https://github.com/ros-visualization/rqt_common_plugins/blob/cc43ec2c46b37b7533b191b7c5258602cc7717a6/package.xml#L55.

> My point was that this ensures that all packages listed in this package.xml are also part of the repos file, so ensuring these packages are part of the various CI.

Since `rqt_common_plugins` doesn't use any of the dependencies at build time it wouldn't fail if e.g.`rqt_topic` wouldn't be available.

There is certainly value to build as many packages as possible across all platforms. But the CI jobs building all packages in the `ros2.repos` are already very long and we simply can't effort adding each and every package we would like to test to it. I also don't see any benefit of adding `rqt_common_plugins` since it doesn't provide any functionality nor does its presence add any more test coverage.

mikaelarguedas:
> I think this will cause problems with some of our CI jobs since not all of the packages listed in the metapackage are in the repos file (e.g. rqt_web). rosdep will fail for example.

This was actually the goal, to catch these untested dependencies.
I was not aware rosdep was being ran on CI jobs. If this is the case this change can have value, otherwise it may be better to close this

mikaelarguedas:
> rqt_common_plugins since it doesn't provide any functionality nor does its presence add any more test coverage

Agreed, if there is no run of rosdep it won't detect the missing deps :+1:. I'll close this PR then.

Regarding the packages that are part of common_plugins and not listed here (e.g. `rqt_action`) is it desired to have them in the repos file ?

dirk-thomas:
> Regarding the packages that are part of common_plugins and not listed here (e.g. `rqt_action`) is it desired to have them in the repos file ?

If they are of general use it would be good to include them to make them available to users using the fat archives (specifically on macOS and Windows where those are the only binaries).

mikaelarguedas:
> If they are of general use

Most of them seem to be of general use but it's pretty subjective.
So I just updated the list to reflect the content of the eloquent meta package and required dependencies, feel free to push directly to this branch to remove any that don't belong.

This compiles on Linux but IDK about the other platforms

ros2.repos
```diff
@@ -39,6 +35,10 @@ repositories:
     type: git
     url: https://github.com/eProsima/foonathan_memory_vendor.git
     version: master
+  eclipse-cyclonedds/cyclonedds:
+    type: git
+    url: https://github.com/eclipse-cyclonedds/cyclonedds.git
+    version: master
```
dirk-thomas: Please avoid moving this entry.

ros2.repos
```diff
@@ -51,10 +51,18 @@ repositories:
...
+    url: https://github.com/ros-perception/image_common.git
+    version: ros2
   ros-perception/laser_geometry:
     type: git
     url: https://github.com/ros-perception/laser_geometry.git
     version: ros2
+  ros-perception/vision_opencv:
+    type: git
+    url: https://github.com/ros-perception/vision_opencv.git
+    version: ros2
```
dirk-thomas: I don't think we have OpenCV available on the build nodes across all platforms. I think that prevents `rqt_image_view` from being added (see #620).

ros2.repos
mikaelarguedas: reverted in https://github.com/ros2/ros2/pull/833/commits/ce64a74a915ac3f04add194636f342e5982a1a61

This is the order outputed by `vcs export`.
Should vcs ignore case when printing results?

ros2.repos
mikaelarguedas: reverted in https://github.com/ros2/ros2/pull/833/commits/4ab9872f5d9e369561fd41d598922c6a895b6b38

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8796/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4616/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7211/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8716/)

dirk-thomas:
Thanks!

PR #837, opened 2019-12-09
Update ros2_tracing to 0.2.11 after ament registration fix

christophebedard:
See `ros2_tracing` issue: https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing/issues/60

This new version also contains some other minor fixes.

rosdistro PR: https://github.com/ros/rosdistro/pull/23257

FYI @dirk-thomas

PR #838, opened 2019-12-09
Update ros2_tracing to 0.2.12

nuclearsandwich:
Getting ahead of releasing a new version to address fixes in CI. There
are other changes to be tested as well however.

Before this is merged, we should update the PR to use the release tag once it is made.

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -42,7 +42,7 @@ repositories:
   micro-ROS/ros_tracing/ros2_tracing:
     type: git
     url: https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing.git
-    version: master
+    version: 0.2.12
   osrf/osrf_pycommon:
     type: git
     url: https://github.com/osrf/osrf_pycommon.git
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8853/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4648/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7249/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8767/)

This will replace #837 since the released version will include those changes.

dirk-thomas:
We decide to not track master but the latest release in the past. If the rational hasn't changed I don't think it should be switched.

christophebedard:
@nuclearsandwich the tag `0.2.12` is ready and the rosdistro PR is up: https://github.com/ros/rosdistro/pull/23259

nuclearsandwich:
> We decide to not track master but the latest release in the past. If the rational hasn't changed I don't think it should be switched

It seems like you did not read the pull request description. I opened the PR (in draft state) so I had somewhere to share CI results with the intent to update to the released tag when it became available.

nuclearsandwich:
The test failures in the above CI are known and accounted for and the specific linter failures in ros2_tracking are resolved with this release.

PR #871, opened 2020-02-18
use eloquent branch from fork

dirk-thomas:
For the bug fix ros2/Fast-RTPS#1.


dirk-thomas:
> though we'll now have to maintain this fork

Only for Eloquent atm which is an already released distro. So it is unlikely that we will do frequent FastRTPS patch releases.

Also using a newer state from the upstream repo is not really an options since they break API / ABI / behavior in patch releases which we can't effort for Eloquent.

PR #892, opened 2020-04-10
[Eloquent] Pin foonathan_memory_vendor to a specific version

mjcarroll:
Attempt to fix recent regression in Eloquent Nightly CI: https://github.com/ros2/build_farmer/issues/270

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -26,7 +26,7 @@ repositories:
   eclipse-cyclonedds/cyclonedds:
     type: git
     url: https://github.com/eclipse-cyclonedds/cyclonedds.git
-    version: 08d9c296f1b569c18188528fd5e8f086e94dd67d 
+    version: 08d9c296f1b569c18188528fd5e8f086e94dd67d
   eProsima/Fast-CDR:
     type: git
     url: https://github.com/eProsima/Fast-CDR.git
@@ -37,7 +37,7 @@ repositories:
     version: ros2-eloquent
   eProsima/foonathan_memory_vendor:
     type: git
-    url: https://github.com/eProsima/foonathan_memory_vendor.git
+    url: https://github.com/mjcarroll/foonathan_memory_vendor.git
     version: master
   micro-ROS/ros_tracing/ros2_tracing:
     type: git


mjcarroll:
[Build Status](https://ci.ros2.org/job/ci_linux/10055/)

cottsay:
I don't think this will be enough. Even 0.5.0 of `foonathan_memory_vendor` is [cloning from master](https://github.com/eProsima/foonathan_memory_vendor/blob/6850cfbaae72d5ff46e6a43f03c203f2875ff898/CMakeLists.txt#L59). The vendor package needs to change.

mjcarroll:
Ah, I see, I misread your summary.

mjcarroll:
Another attempt with pinning back the `foonathan/memory` version in the `vendor` package (https://github.com/mjcarroll/foonathan_memory_vendor/commit/9fbf72fa918e839f8424b827f487c507cafebcf1).

[Build Status](https://ci.ros2.org/job/ci_linux/10057/)

I don't know if this PR makes sense unless we want to additionally pin the vendor package to a revision that incorporates that change.

cottsay:
If we're going to maintain a fork until eProsima releases a solution, I think we should probably fork into `ros2`.

That said, I think this is the right thing to do to unblock `Eci` builds right away. Eventually we should make sure that all of the ros2 distros, bot the development and `-release` branches, point to a version of `foonathan_memory_vendor` that pins `foonathan/memory`.

mjcarroll:
> If we're going to maintain a fork until eProsima releases a solution, I think we should probably fork into ros2.

I'm hoping that this lands quickly enough so that the fork is unnecessary.  I forked to my own mainly to get the PR up and run it through CI.

nuclearsandwich:
> > If we're going to maintain a fork until eProsima releases a solution, I think we should probably fork into ros2.

Is the PR on foonathan_memory_vendor open? I'm a maintainer of the package and can review and potentially merge as long as it doesn't cause any break's on eProsima's infrastructure.

cottsay:
I believe that this was resolved by eProsima/foonathan_memory_vendor#24. That said, we should create and release 0.4.1 and 0.5.1 into Dashing and Eloquent so that source builds in the `*-release` branches of `ros2` get the fix.

mjcarroll:
Opened an issue to cover that: https://github.com/eProsima/foonathan_memory_vendor/issues/25

Issue #898, opened 2020-04-15
Track released version of ros2_tracing


nuclearsandwich:
In an upcoming PR I'll pin a commit of ros2_tracing which fixes a deprecation warning currently affecting CI buids. This issue tracks lifting that pin and returning to the official release track of ros2_tracing.

rotu:
As mentioned in #899, if we're going to include it in ros2, we may as well use master. Despite @dirk-thomas, just because it lives outside of github doesn't mean the `master` branch is less stable or suitable than that of other projects.

dirk-thomas:
@rotu Please don't ignore feedback you get on other tickets - I commented twice on the referenced ticket.

There is a rational *why* we chose to use a tag in the past. Just because you feel differently isn't sufficient to change the decision. You need to bring up good enough arguments why the rational should be changed. And also consider what the impact of the proposed change is (e.g. if something break we don't have access to that repo is a pretty big deal). Changes from this third-party repo are gated on a pull request which explicitly levels the version and is usually with a set of CI builds showing that it is regression free.

rotu:
I did read your feedback and I discovered this issue, which is, I think a better place to present the feedback. You presented two rationales:

* "changes may regress the build" and "you don't have write access to the repo"
    *If something breaks, you can hit up the maintainers to fix it swiftly or pin back the repos file to a working version.
* "This repo doesn't have CI"
    * Adding CI to this repo is something that can be done. https://docs.gitlab.com/ee/integration/jenkins.html Or it can be run on each MR manually like the current process.

Here are the rationales for designating and using a stable branch instead:

* Upgrade path is automated and more obvious. It's silly to have to check out a new ros2.repos just because other ROS2 projects introduce compatibility breaks and/or deprecations or because this project makes improvements. By using a branch, `git pull` or `vcs pull` does that automatically
* Following a branch shows trust in the developer. ROS2 encourages widespread contribution and participation. By not allowing projects we use to specify their own trajectory, you're saying "we don't trust you to release stable, high quality code". It betrays the community principles.

nuclearsandwich:
<details><summary>Protocol sidebar</summary>

> I did read your feedback and I discovered this issue, which is, I think a better place to present the feedback.

[...]

PR #899, opened 2020-04-15
Bump to the latest commit on ros2_tracing to fix deprecation.

nuclearsandwich:
Since the release of ros2_tracing including this change is delayed we'll
temporarily pin the commit which addresses the deprecation warning.

There are a substantial number of changes since the last release ([0.3.0...master](https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing/-/compare/0.3.0...master)) so this may not go down easy.

This PR "causes" #898 which will be resolved when we're back on a release channel.

Signed-off-by: Steven! Ragnarök <steven@nuclearsandwich.com>

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -42,7 +42,7 @@ repositories:
   micro-ROS/ros_tracing/ros2_tracing:
     type: git
     url: https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing.git
-    version: 0.3.0
+    version: c3ea84b4b98a99d0a7ba3ed94846e1aaec8a8d13
   osrf/osrf_pycommon:
     type: git
     url: https://github.com/osrf/osrf_pycommon.git


nuclearsandwich:
First verify that this builds and resolves the CMake warning by building up-to tracetools.

[Build Status](https://ci.ros2.org/job/ci_linux/10230/)

nuclearsandwich:
I tried to scope this but the upstream and downstream was 244/298 packages and using packages-select with it overflows the command length buffer locally and I can't imagine CI would enjoy it. So here's full CI with the new tracetools changes

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10233/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5780/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8321/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10120/)

nuclearsandwich:
The windows build hung during tests (ros2/build_farmer#248) but only after completing a full build.

Security failures with Connext on macOS are related to ros2/system_tests#409
cli tests in ros2cli packages were affected by ros2/rclpy#536 and other failing tests are flakey with work to improve them ongoing.

rotu:
Why is the version a commit hash instead of the `foxy` or `master` branch? Are these branches too unstable?

christophebedard:
> Why is the version a commit hash instead of the `foxy` or `master` branch? Are these branches too unstable?

As @nuclearsandwich mentioned, I have to delay the Foxy release. I want to include some more things for the quality declaration, but I personally haven't had time yet (I blame school work :stuck_out_tongue:)

rotu:
Hmmm... Even if it's not release-ready, I'd still rather use `master` so I can `vcs pull` to check for changes as they roll in, so long as it doesn't tend to break the build.

rotu:
Sure, if master is unsuitably unstable that it may regress the build, can we maybe create a *new* branch for ros2.repos to track that is slower moving so as not to regress the build? Your rationale applies to *every* package in ros2.repos.

dirk-thomas:
> Your rationale applies to every package in ros2.repos.

The significant difference that we don't have write ccess to `tracetools`. Any third party repositories like this we prefer to use tags / hashes or branches (if stable) over `master`.

> we maybe create a new ros2_tracing branch for ros2.repos to track that is slower moving so as not to regress the build?

Upstream is planning to tag a new version soon (https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing/-/merge_requests/163#note_321456164) and then we will use that tag (as we have done before).

rotu:
> Any third party repositories like this we prefer to use tags / hashes or branches (if stable) over `master`.

`master` does seem to be such a a fairly stable branch, and `foxy` seems to be a fine candidate as well. And if anything, `master` seems to adapt more *slowly* than many core ROS2 packages, which is not a problem that tags can solve. (in the latest case, the version bump was to fix a build problem, which should have been done with just a `vcs pull` rather than a `vcs import`)

PR #894, opened 2020-04-13
Switch cyclone from master to its 0.5.x release branch

rotu:

PR #905, opened 2020-04-22
Use cyclonedds master branch for ros2 master

eboasson:
#894 changed cyclonedds in `master:ros2.repos` to use the releases/0.5.x branch, but I think that was a mistake. (It basically never registered that Foxy was being changed as well, which is my fault, not @rotu's.) In the current state, Foxy can't pick up changes in cyclonedds, which then creates a problem for, e.g., https://github.com/ros2/rmw_cyclonedds/pull/145 that relies on a new function.

This PR reverts #894 by switching it back to master, which I think is more sensible than the two alternatives I see:
* backporting that new function onto cyclonedds' releases/0.5.x branch for no other reason than keeping Foxy pinned on that branch, and so pushing "unnecessary" changes to a release branch;
* creating a releases/0.6.x branch right now for cyclone, even though that cuts the branch at a point where there are definitely going to be a few PRs needed to get it to point where I could treat it as a release candidate (stuff like doc updates and paperwork).

As feature freeze for Foxy must be about now, I can imagine that this PR is really bad timing. The other options are possible if there is a consensus that either of those is a better choice.

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -26,7 +26,7 @@ repositories:
   eclipse-cyclonedds/cyclonedds:
     type: git
     url: https://github.com/eclipse-cyclonedds/cyclonedds.git
-    version: releases/0.5.x
+    version: master
   eProsima/Fast-CDR:
     type: git
     url: https://github.com/eProsima/Fast-CDR.git


dirk-thomas:
This is fine. In all other ROS 2 repos we create release branches later, too. Usually in the week before the release.

rotu:
Oof! That was totally my mistake!!! I had meant to make the PR against Dashing, which still accidentally targets cyclonedds master https://github.com/ros2/ros2/blob/dashing/ros2.repos

PR #996, opened 2020-07-31
Add pybind11_vendor

mabelzhang:
Pending merge of https://github.com/ros2/pybind11_vendor/pull/2
Needed for rosbag2 Python API https://github.com/ros2/rosbag2/pull/308

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -231,6 +231,10 @@ repositories:
     type: git
     url: https://github.com/ros2/performance_test_fixture.git
     version: main
+  ros2/pybind11_vendor:
+    type: git
+    url: https://github.com/ros2/pybind11_vendor.git
+    version: master
   ros2/python_cmake_module:
     type: git
     url: https://github.com/ros2/python_cmake_module.git


dirk-thomas:
This change resulted in a CMake warning on Windows: e.g. https://ci.ros2.org/view/nightly/job/nightly_win_rel/1649/msbuild/new/

@mabelzhang Please resolve asap. The same warning was also present in the CI builds on the referenced PR which raises the question why this was merged anyway?

hidmic:
BTW you can find one correct solution to that problem [here](https://github.com/ros2/mimick_vendor/blob/d6a3d7efe0af01d09341bf9f753369c8d686c661/CMakeLists.txt#L22-L29). I've seen some vendor packages suppressing warnings instead.

mabelzhang:
Hmm that warning was gone in the latest CI builds in the two PRs above, which had no warnings for any of the platforms, so I thought it went away.

Thank you @hidmic, I will try the solution.

PR #1005, opened 2020-08-13
Missing ros-visualization/tango_icon_vendor package on foxy branch

guiboro83:

PR #1011, opened 2020-08-15
add missing tango_icons_vendor

mikaelarguedas:
Otherwise attempting to build Foxy following the [installation instructions](https://index.ros.org//doc/ros2/Installation/Foxy/Linux-Development-Setup)  fails as `rqt_gui` in foxy depends on `tango_icons_vendor` (a similar fix is likely needed on the foxy-release branch)

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -131,6 +131,10 @@ repositories:
     type: git
     url: https://github.com/ros-visualization/rqt_topic.git
     version: dashing-devel
+  ros-visualization/tango_icons_vendor:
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
A PR with the same change is already available: https://github.com/ros2/ros2/pull/1005/

jacobperron:
We should certainly add this package to the repos file, but I can't reproduce the build failure. I'm surprised that this didn't come up when building the latest Foxy patch release :thinking:

PR #1113, opened 2021-03-23
Upgrade Fast DDS to 2.1.x

EduPonz:
Signed-off-by: EduPonz <eduardoponz@eprosima.com>

PR #1114, opened 2021-03-23
Upgrade Fast DDS to v2.1.1

EduPonz:
Signed-off-by: EduPonz <eduardoponz@eprosima.com>

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -34,7 +34,7 @@ repositories:
   eProsima/Fast-DDS:
     type: git
     url: https://github.com/eProsima/Fast-DDS.git
-    version: v2.0.2
+    version: v2.1.0
   eProsima/foonathan_memory_vendor:
     type: git
     url: https://github.com/eProsima/foonathan_memory_vendor.git
--- a/ros2.repos
+++ b/ros2.repos
@@ -34,7 +34,7 @@ repositories:
   eProsima/Fast-DDS:
     type: git
     url: https://github.com/eProsima/Fast-DDS.git
-    version: v2.1.0
+    version: v2.1.1
   eProsima/foonathan_memory_vendor:
     type: git
     url: https://github.com/eProsima/foonathan_memory_vendor.git


jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14304/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9093/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11990/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14409/)

mikaelarguedas:
@clalancette Security is currently broken on Foxy without this change, is it possible to get this in or does this involve significant overhead on the ROS 2 side ?

---

This is the [second time](https://github.com/ros2/sros2/issues/252) Security gets broken in Foxy because of Fast-RTPS changes. Any strategy we can put in place to avoid having breaking changes landing in LTS distros in the future ?

clalancette:
> @clalancette Security is currently broken on Foxy without this change, is it possible to get this in or does this involve significant overhead on the ROS 2 side ?

Can you explain a bit more about why security is broken?  As far as I know the patch that I applied in https://github.com/ros2-gbp/fastrtps-release/pull/12 to fix it last time is still there.

mikaelarguedas:
yes the patch is still here. Now it is broken because of https://github.com/ros2/rmw_fastrtps/issues/522#issuecomment-816568387
(both being the result of bumping Fast-RTPS from 2.0.0 to 2.0.2 after Foxy was out)

clalancette:
> yes the patch is still here. Now it is broken because of [ros2/rmw_fastrtps#522 (comment)](https://github.com/ros2/rmw_fastrtps/issues/522#issuecomment-816568387)
> (both being the result of bumping Fast-RTPS from 2.0.0 to 2.0.2 after Foxy was out)

I don't have a lot of great solutions here.  We can pin down the core repositories a lot more, but it means that it is hard to get fixes out to users.  The other option is to get more community testing of https://packages.ros.org/ros2-testing, but that has not worked in the past.  I'm open to other suggestions.

MiguelCompany:
@clalancette @mikaelarguedas 

The issue, in this case, was only reproduced when there are several subscriptions for the same protected topic. Adding a test for that would be convenient.

MiguelCompany:
As an additional note, having jobs similar to [this one](https://build.ros2.org/view/Fdev/job/Fdev__cyclonedds__ubuntu_focal_amd64/) for Fast DDS would help detect issues earlier

nuclearsandwich:
> As an additional note, having jobs similar to [this one](https://build.ros2.org/view/Fdev/job/Fdev__cyclonedds__ubuntu_focal_amd64/) for Fast DDS would help detect issues earlier

Ask and you shall receive. https://github.com/ros/rosdistro/pull/29124
As mentioned in the above PR. Because the SECURITY option is not on by default it won't actually be built in our devel jobs. I am not sure whether it's possible or preferable to change the CMake configuration for future releases so that SECURITY is on by default if OpenSSL is found but that would cut down on the patching we do at release time and it would ensure that devel jobs build and run with security tests.

nuclearsandwich:
I also updated the source entry for Foxy to match this proposal. If this is merged then  https://github.com/ros/rosdistro/pull/29125 should be as well.

MiguelCompany:
Thanks @nuclearsandwich!

> SECURITY option is not on by default

I thought one of the goals of [this repo](https://github.com/colcon/colcon-metadata-repository) was precisely to always configure Fast DDS with security. Could it be enabled in some way?

ros-discourse:

https://discourse.ros.org/t/discovery-problems-with-the-discovery-server/19900/6

mikaelarguedas:
>  We can pin down the core repositories a lot more, but it means that it is hard to get fixes out to users

I'm all for getting critical bug fixes to users :+1:
Would there be a way to enforce or at least strongly advise that patch releases within an LTS distro should have *only* critical bug fixes ?
The jumps from 2.0.0 to 2.0.2 was *massive*: https://github.com/ros2-gbp/fastrtps-release/compare/upstream/2.0.0...upstream/2.0.2
And the suggestion here to fix the bugs introduced by that patch release is to jump a minor release ahead. Which seems very odd to me, and that's why I want to raise the concern. What is the current policy regarding very low level package changes?

---

> The issue, in this case, was only reproduced when there are several subscriptions for the same protected topic. Adding a test for that would be convenient.

Thanks that would be a nice addition indeed. https://github.com/ros2/sros2/issues/261 has been update to track that on the ROS side. Can you confirm if this kind of scenario is covered by Fast DDS automated testing ?

MiguelCompany:
> Can you confirm if this kind of scenario is covered by Fast DDS automated testing ?

The tests were added at the same time as the fix. See [here](https://github.com/eProsima/Fast-DDS/pull/1645/files#diff-6cad93b063fb6d463ba627e32b899eb7d83feee8624ab1d539fc7bcb7de45a6aR87).

EduPonz:
> Which seems very odd to me, and that's why I want to raise the concern. What is the current policy regarding very low level package changes?

AFAIK, the policy is that changes should be API and ABI compatible. The reason why this is a minor release instead of a micro is that Fast DDS has API beyond what ROS 2 uses and exposes to its users. In this case, v2.0.2 and v2.1.1 are API and ABI compatible for ROS 2 usage.

mikaelarguedas:
> The tests were added at the same time as the fix. See here.

> AFAIK, the policy is that changes should be API and ABI compatible

Thanks you both for clarifying :+1:

clalancette:
The foxy-release branch will get automatically updated once we do a Foxy patch release.  So I'm going to close this one out in favor of #1113 .

PR #1133, opened 2021-04-22
Pin cyclonedds temporarily to avoid CI failures

ivanpauno:
CI builds are failing after https://github.com/eclipse-cyclonedds/cyclonedds/pull/768, which requires https://github.com/ros2/rmw_cyclonedds/pull/306.

I think we don't want to introduce API breaks at this point.
Pinning cyclonedds temporaly will put out the CI fire :fire: :fire_engine: until that commit can be reverted.

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -30,7 +30,7 @@ repositories:
   eclipse-cyclonedds/cyclonedds:
     type: git
     url: https://github.com/eclipse-cyclonedds/cyclonedds.git
-    version: 45a9ac292a1863df7d79f7d1919d19b09f1f9807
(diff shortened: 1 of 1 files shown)


cottsay (approved):

eboasson:
eclipse-cyclonedds/cyclonedds#768 has been reverted

clalancette:
> [eclipse-cyclonedds/cyclonedds#768](https://github.com/eclipse-cyclonedds/cyclonedds/pull/768) has been reverted

Given that, I'm going to close this out.  Thanks everyone.

PR #1152, opened 2021-06-09
Update release versions for the final patch release.

nuclearsandwich:
Some repository names have changed and other repositories have been
merged from their ROS 2 fork back into the upstream repository.

Additionally updates the release versions for this patch release.

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -34,7 +34,7 @@ repositories:
   eProsima/Fast-DDS:
     type: git
     url: https://github.com/eProsima/Fast-DDS.git
-    version: v1.8.2
+    version: v1.8.4
   osrf/osrf_pycommon:
     type: git
     url: https://github.com/osrf/osrf_pycommon.git
@@ -62,7 +62,7 @@ repositories:
   ros-visualization/rqt:
     type: git
     url: https://github.com/ros-visualization/rqt.git
-    version: 1.0.5
+    version: 1.0.7
   ros-visualization/rqt_console:
     type: git
     url: https://github.com/ros-visualization/rqt_console.git
@@ -134,7 +134,7 @@ repositories:
   ros2/common_interfaces:
     type: git
     url: https://github.com/ros2/common_interfaces.git
-    version: 0.7.0
+    version: 0.7.1
   ros2/console_bridge_vendor:
     type: git
     url: https://github.com/ros2/console_bridge_vendor.git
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14736/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9487/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12401/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14882/)

ros2.repos
```diff
@@ -31,10 +31,10 @@ repositories:
     type: git
     url: https://github.com/eProsima/Fast-CDR.git
     version: v1.0.13
-  eProsima/Fast-RTPS:
+  eProsima/Fast-DDS:
```
clalancette, line 34: I'm fine with changing the URL to the up-to-date version, but should we change the checkout location?  My thinking is that this will cause some (minor) headaches to anyone who is building from source.  Since this is the last release of Dashing, I kind of think we should leave this alone.  The same goes for `robot_state_publisher` and `urdfdom` below.

nuclearsandwich: I guess changing the checkout location isn't required. I just did it to match. I updated the urls because I use them in keys for the script I have to update the repos file from a rosinstall_generator --upstream and they've been updated in dashing/distribution.yaml.

clalancette: > I guess changing the checkout location isn't required. I just did it to match. I updated the urls because I use them in keys for the script I have to update the repos file from a rosinstall_generator --upstream and they've been updated in dashing/distribution.yaml.

Yeah, understood.  It's just that if a user has a checkout of the previous ros2.repos file, then does a `wget` of this new one and a `vcs import`, they'll now have duplicate repositories that colcon will be cranky about.

But it is admittedly a minor point.  I'm going to go ahead and approve, and I'll let you decide whether that use-case is important enough to warrant a change here.

clalancette (approved):

ros2.repos
nuclearsandwich: I decided to preserve the checkout paths for these repositories even if the immedate reason was to avoid having to re-sort them based on the directory changes.

nuclearsandwich:
The issues shown in CI above are all accounted for as pre-existing issues in Dashing.

wjwwood (approved):

ahcorde (approved):

cottsay (approved):

jacobperron (approved):
:runner: :checkered_flag:

tfoote (approved):

PR #1171, opened 2021-08-24
Remove non-default RMW from minimally scoped repos list.

nuclearsandwich:
For this project we're tightly limiting the scope of packages we build and test. In this case we don't need to include the non-default rmw provider packages in the build space.


ros2.repos
```diff
@@ -179,18 +167,10 @@ repositories:
     type: git
     url: https://github.com/ros2/rmw_connextdds.git
     version: master
-  rmw_cyclonedds:
```
mjeronimo, line 182: Removing this would leave the 'cyclonedds' package and the 'rmw_connextdds' package. Shouldn't it be that 'rmw_cyclonedds' is required and 'rmw_connextdds' is removed?

nuclearsandwich: Thank you. Since the rmw_connextdds package rename I am constantly getting these backwards.

nuclearsandwich: The last commit has been rewritten as 7cf9059 which removes rmw_connextdds rather than rmw_cyclonedds

nuclearsandwich:
[Build Status](https://ci.ros2.org/job/ci_linux/15105/) reference build on Linux running no tests just making sure that the build is coherent.

PR #1187, opened 2021-09-15
Update the ros2.repos file for Space ROS

mjeronimo:
This was generated using the latest Space ROS package list, with a few changes and additions:

* Updated the ament_lint branch to use 'spaceros' instead of 'master'

* Added ros2cli for convenient acces to the ros2 utility. In the future we may have a finer, layered approach to the package lists, but for now it will be convenient to have the ros2 utility available.

* Added rosidl_runtime_py since it is required by some of the ros2 sub-commands.

Signed-off-by: Michael Jeronimo <michael.jeronimo@openrobotics.org>

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -39,18 +39,6 @@ repositories:
     type: git
     url: https://github.com/ros2/eigen3_cmake_module.git
     version: master
-  fastcdr:
-    type: git
-    url: https://github.com/eProsima/Fast-CDR.git
-    version: v1.0.13
-  fastrtps:
-    type: git
-    url: https://github.com/eProsima/Fast-DDS.git
-    version: 2.3.x
-  foonathan_memory_vendor:
-    type: git
-    url: https://github.com/eProsima/foonathan_memory_vendor.git
-    version: master
   geometry2:
     type: git
     url: https://github.com/ros2/geometry2.git
@@ -155,10 +143,6 @@ repositories:
     type: git
     url: https://github.com/ros2/rmw.git
     version: master
-  rmw_connextdds:
-    type: git
-    url: https://github.com/ros2/rmw_connextdds.git
-    version: master
   rmw_cyclonedds:
     type: git
     url: https://github.com/ros2/rmw_cyclonedds.git
@@ -166,10 +150,6 @@ repositories:
   rmw_dds_common:
     type: git
     url: https://github.com/ros2/rmw_dds_common.git
-    version: master
-  rmw_fastrtps:
-    type: git
-    url: https://github.com/ros2/rmw_fastrtps.git
     version: master
   rmw_implementation:
     type: git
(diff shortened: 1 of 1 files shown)


ros2.repos
```diff
@@ -155,14 +151,14 @@ repositories:
...
-    url: https://github.com/ros2/realtime_support.git
-    version: master
   rmw:
     type: git
     url: https://github.com/ros2/rmw.git
     version: master
+  rmw_connextdds:
+    type: git
+    url: https://github.com/ros2/rmw_connextdds.git
+    version: master
```
wjwwood, line 161: I don't think we want connextdds and its dependencies.

ros2.repos
```diff
@@ -171,6 +167,10 @@ repositories:
     type: git
     url: https://github.com/ros2/rmw_dds_common.git
     version: master
+  rmw_fastrtps:
+    type: git
+    url: https://github.com/ros2/rmw_fastrtps.git
+    version: master
```
wjwwood, line 173: I don't think we want rmw_fastrps and its dependencies.

wjwwood (changes requested):

wjwwood:
I don't see `ros2cli` or `rosidl_runtime_py` in the diff, is that expected?

ros2.repos
mjeronimo: Ok. Connext and FastRTP)S are in the spaceros-pkgs.txt (added by Simon). Ok to delete there?

mjeronimo:
Regarding ros2cli and rosidl_runtime_py: They weren't in spaceros-pkgs.txt from which I generated the ros2.repos file. So, in that sense, they were "added." However, they were already in the previous ros2.repos file on the spaceros branch, which is a much larger set of packages (and not generated from the package list). So, relative to that, they were already present and won't show up in the diffs.

wjwwood:
That's a good question, maybe @shonigmann can comment. I thought we'd decided to remove fastrtps and connext, but looking again I think we just said that's what we wanted to do but didn't prune the list.

wjwwood:
> Updated the ament_lint branch to use 'spaceros' instead of 'master'

I don't see this in the diff either?

mjeronimo:
Same issue. My comments were probably misleading because they were relative to the input spaceros-pkgs list. The new ros2.repos was generated from this list using 'rolling' and so required an update to the ament_lint branch name. However, the ament_lint branch was already 'spaceros' in the existing ros2.repos file against which the generated file is diffed.

wjwwood:
Got it.

wjwwood (approved):
I guess this ok then, we'll just have to update it later as the source material changes.

shonigmann:
> That's a good question, maybe @shonigmann can comment. I thought we'd decided to remove fastrtps and connext, but looking again I think we just said that's what we wanted to do but didn't prune the list.

That's correct - the .repos file was just generated using rosinstall_generator to include all dependencies, as is. I didn't go through and prune the list because I thought it may cause build issues. In any case, the vnv focus list should still represent the packages who's output we are interested in looking at.

mjeronimo:
Ok, I'll prune those from this repos file.

PR #1188, opened 2021-09-21
Add ament_ikos to the repos file for Space ROS

mjeronimo:
Signed-off-by: Michael Jeronimo <michael.jeronimo@openrobotics.org>


wjwwood (approved):
lgtm, had a discussion offline about how ikos is provided as a dependency.

PR #1261, opened 2022-04-25
[humble] Update repos file

audrow:
This PR updates the repos file using the [`ros2-branching`](https://github.com/audrow/ros2-branching) repo for most of its changes and then adds some manual changes.

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -78,7 +78,7 @@ repositories:
   ros-tooling/keyboard_handler:
     type: git
     url: https://github.com/ros-tooling/keyboard_handler.git
-    version: main
+    version: humble
   ros-tooling/libstatistics_collector:
     type: git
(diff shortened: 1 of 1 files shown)


ros2.repos
```diff
@@ -2,39 +2,39 @@ repositories:
...
   eProsima/Fast-CDR:
     type: git
     url: https://github.com/eProsima/Fast-CDR.git
-    version: master
+    version: v1.0.24
   eProsima/Fast-DDS:
     type: git
     url: https://github.com/eProsima/Fast-DDS.git
-    version: master
+    version: v2.6.0
```
clalancette, line 37: Not quite.  For Fast-DDS, we use the branch, not the tag, so this should be:

```suggestion
    version: 2.6.x
```

clalancette (changes requested):
One change, then this is good to go.

clalancette (approved):

PR #1264, opened 2022-05-04
Remove rqt_top from ros2.repos.

clalancette:
As it stands, it only shows a list of nodes and no other information.
To make it actually useful, we'd have to do a bunch more work,
so remove it until someone steps up to do that work.

I'll also backport this change to Humble once it is approved.


audrow (approved):

clalancette:
@Mergifyio backport humble

PR #1267, opened 2022-05-06
Add scripts to generate a repos file.

nuclearsandwich:
The repos file in this repository was initially created with rosinstall_generator.
We've manually updated it after changes in ROS 2 and in Space ROS-specific branches and projects.

I'd like to get to a point where we have specified via scripted logic the update process from ROS 2 and the incorporation of Space ROS modifications so that we can use the repos file as an input artifact to other builds in such a way that changes to upstream or Space ROS will generate an updated repos file sufficient to break cache on a docker build with the repos file as an input.

This is draft work for now. I usually like to work on generators until they generate 1:1 the existing file but given changes in upstream that may not be entirely achievable.


nuclearsandwich:
Wrong target repo :facepalm:

PR #1274, opened 2022-05-19
Set Humble release versions

audrow:
This is the result of running the following command on the Humble repos cloned from the repos file on the `humble` branch.
```bash
vcs export –exact-with-tags src
```

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -34,7 +34,7 @@ repositories:
   eProsima/Fast-DDS:
     type: git
     url: https://github.com/eProsima/Fast-DDS.git
-    version: e6e7eac5a642f606fdaa40a0a2a16617ae9384da
+    version: 2.6.x
   eProsima/foonathan_memory_vendor:
     type: git
     url: https://github.com/eProsima/foonathan_memory_vendor.git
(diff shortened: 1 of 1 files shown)


clalancette (changes requested):
This looks generally good, with a few exceptions:

1.  We should switch Fast-DDS to the tag that we are using for the binary packages, `v2.6.0`
2. We should switch cyclonedds to the tag that we are using for the binary packages, `v0.9.0`
3. We should switch iceoryx to the tag that we are using for the binary packages, `v2.0.2`.

audrow:
> 2\. We should switch cyclonedds to the tag that we are using for the binary packages, `v0.9.0`

Shouldn't cyclonedds go to `0.9.0` (no `v`)? https://github.com/eclipse-cyclonedds/cyclonedds/releases/tag/0.9.0

clalancette:
> Shouldn't cyclonedds go to `0.9.0` (no `v`), or am I missing something? https://github.com/eclipse-cyclonedds/cyclonedds/releases/tag/0.9.0. The other two I can find matching tags for.

Yes!  Sorry, typo on my part.

clalancette (approved):

PR #1320, opened 2022-09-13
[galactic] Update source branches for rqt_action and rqt_console

jacobperron:
I also noticed that we're tracking the `ros2` branch of [rqt_bag](https://github.com/ros-visualization/rqt_bag) for both Rolling and Galactic (while Humble is tracking it's own, older branch). I'm not sure what to do in that case. Open to suggestions.

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -146,7 +146,7 @@ repositories:
   ros-visualization/tango_icons_vendor:
     type: git
     url: https://github.com/ros-visualization/tango_icons_vendor.git
-    version: humble
+    version: galactic
   ros/class_loader:
     type: git
     url: https://github.com/ros/class_loader.git
--- a/ros2.repos
+++ b/ros2.repos
@@ -146,7 +146,7 @@ repositories:
   ros-visualization/tango_icons_vendor:
     type: git
     url: https://github.com/ros-visualization/tango_icons_vendor.git
-    version: master
+    version: humble
   ros/class_loader:
     type: git
     url: https://github.com/ros/class_loader.git
--- a/ros2.repos
+++ b/ros2.repos
@@ -90,7 +90,7 @@ repositories:
   ros-visualization/rqt_action:
     type: git
     url: https://github.com/ros-visualization/rqt_action.git
-    version: ros2
+    version: galactic
   ros-visualization/rqt_bag:
     type: git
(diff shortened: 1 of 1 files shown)


jacobperron:
I also updated `tango_icons_vendor` to target the `humble` branch, which is where the latest Galactic release was made from.

ros2.repos
```diff
@@ -90,15 +90,15 @@ repositories:
   ros-visualization/rqt_action:
     type: git
     url: https://github.com/ros-visualization/rqt_action.git
-    version: ros2
+    version: galactic
```
clalancette, line 93: I see you've also opened https://github.com/ros/rosdistro/pull/34452 for this.  :+1:

ros2.repos
```diff
@@ -90,15 +90,15 @@ repositories:
...
   ros-visualization/rqt_bag:
     type: git
     url: https://github.com/ros-visualization/rqt_bag.git
     version: ros2
   ros-visualization/rqt_console:
     type: git
     url: https://github.com/ros-visualization/rqt_console.git
-    version: ros2
+    version: galactic
```
clalancette, line 101: I see you've also opened https://github.com/ros/rosdistro/pull/34451 for this.  :+1:

ros2.repos
```diff
@@ -146,7 +146,7 @@ repositories:
   ros-visualization/tango_icons_vendor:
     type: git
     url: https://github.com/ros-visualization/tango_icons_vendor.git
-    version: master
+    version: humble
```
clalancette, line 149: This one is pretty weird; we are referencing a branch *ahead* of Galactic.  Honestly, we are pretty unlikely to change this package during Galactic's lifetime, so I'll suggest we just branch off a `galactic` branch that is identical to the `humble` branch, and just use that.  It will make it easier to understand.

clalancette (changes requested):
One thing I think we should change, then this is good to go.

ros2.repos
jacobperron: Done [e285c26](https://github.com/ros2/ros2/pull/1320/commits/e285c261aed0f77e3e0d8d452a6148b6d265d64a)

jacobperron: rosdistro: https://github.com/ros/rosdistro/pull/34473

clalancette (approved):

PR #1397, opened 2023-03-16
Foxy release versions 2023-03-16

quarkytale:
Foxy patch release 10

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18252/)
* Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/12795/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18927/)
* CentOS [Build Status](https://ci.ros2.org/view/All/job/ci_linux-rhel/347/)

Packaging:
* Linux [Build Status](https://ci.ros2.org/job/ci_packaging_linux/542/)
* Windows [Build Status](https://ci.ros2.org/job/ci_packaging_windows/223/)
* Windows Debug [Build Status](https://ci.ros2.org/job/ci_packaging_windows/224/)
* CentOS [Build Status](https://ci.ros2.org/job/ci_packaging_linux-rhel/40/)

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -10,7 +10,7 @@ repositories:
   ament/ament_lint:
     type: git
     url: https://github.com/ament/ament_lint.git
-    version: 0.9.7
+    version: 0.9.8
   ament/ament_package:
     type: git
     url: https://github.com/ament/ament_package.git
@@ -38,7 +38,7 @@ repositories:
   eProsima/foonathan_memory_vendor:
     type: git
     url: https://github.com/eProsima/foonathan_memory_vendor.git
-    version: v1.2.1
+    version: v1.3.0
   eclipse-cyclonedds/cyclonedds:
     type: git
     url: https://github.com/eclipse-cyclonedds/cyclonedds.git
@@ -62,7 +62,7 @@ repositories:
   ros-tooling/libstatistics_collector:
     type: git
     url: https://github.com/ros-tooling/libstatistics_collector.git
-    version: 1.0.1
+    version: 1.0.2
   ros-visualization/interactive_markers:
     type: git
     url: https://github.com/ros-visualization/interactive_markers.git
@@ -82,7 +82,7 @@ repositories:
   ros-visualization/rqt_action:
     type: git
     url: https://github.com/ros-visualization/rqt_action.git
-    version: 1.0.1
+    version: 1.0.2
   ros-visualization/rqt_console:
     type: git
     url: https://github.com/ros-visualization/rqt_console.git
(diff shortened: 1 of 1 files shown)


ros2.repos
```diff
@@ -38,7 +38,7 @@ repositories:
   eProsima/foonathan_memory_vendor:
```
clalancette, line 38: On the line just prior to this one, we also need to update Fast-DDS to v2.1.3.

ros2.repos
```diff
@@ -38,7 +38,7 @@ repositories:
...
     type: git
     url: https://github.com/eProsima/foonathan_memory_vendor.git
-    version: v1.2.1
+    version: v1.3.0
```
clalancette, line 41: This is not the version we currently have released in https://github.com/ros/rosdistro/blob/cbd57a5f617afd0b652b4579ed14a83ebf0c8367/foxy/distribution.yaml#L1514 .

```suggestion
    version: v1.2.0
```

ros2.repos
```diff
@@ -202,7 +202,7 @@ repositories:
   ros2/geometry2:
```
clalancette, line 202: I can't comment on the correct line due to GitHub limitations, but tango_icon_vendors is incorrect.  It should be 0.1.0: https://github.com/ros/rosdistro/blob/cbd57a5f617afd0b652b4579ed14a83ebf0c8367/foxy/distribution.yaml#L6864

ros2.repos
```diff
@@ -226,11 +226,11 @@ repositories:
   ros2/orocos_kinematics_dynamics:
     type: git
     url: https://github.com/ros2/orocos_kinematics_dynamics.git
-    version: 3.3.4
+    version: 3.3.5
   ros2/performance_test_fixture:
     type: git
     url: https://github.com/ros2/performance_test_fixture.git
-    version: 0.0.9
+    version: 0.1.0
```
clalancette, line 233: This isn't what we currently have released in https://github.com/ros/rosdistro/blob/cbd57a5f617afd0b652b4579ed14a83ebf0c8367/foxy/distribution.yaml#L3826 , so we should go back to the old one:

```suggestion
    version: 0.0.9
```

clalancette (changes requested):

ros2.repos
quarkytale: It's weird as I exported tags it added a commit id to the file instead of a version number, so I reverted it to the previous one. Was going to ping you for confirmation, thanks for updating!

clalancette (changes requested):
We also need an update for rosbag2 to 0.3.10.  Otherwise, this looks good.

clalancette (approved):
The changes here look good to me with green CI.

As an FYI, when you are running the CI jobs for RHEL, you need to set the CI_EL_RELEASE field explicity to `8` for Foxy.

PR #1404, opened 2023-04-06
[Humble] Backport CycloneDDS 0.10.x release.

bperseghetti:
Update CycloneDDS to 0.10.x release.

This allows for use of `presence_required` for transient configured networks.

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -42,7 +42,7 @@ repositories:
   eclipse-cyclonedds/cyclonedds:
     type: git
     url: https://github.com/eclipse-cyclonedds/cyclonedds.git
-    version: releases/0.9.x
+    version: releases/0.10.x
   eclipse-iceoryx/iceoryx:
     type: git
     url: https://github.com/eclipse-iceoryx/iceoryx.git


nuclearsandwich:
Here's an initial run of comprehensive CI, I did not disable linter tests, which I used to do when testing stable distros due to linter drift but I think that we've mostly moved to using Ubuntu-distributed linters so that may be less of an issue now. We'll find out.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18459/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13005/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19152/)

clalancette:
Regardless of the outcome, I feel like we should hold off on backporting this to a stable distro for a while.  After all, we only updated Rolling this morning.

audrow:
> Regardless of the outcome, I feel like we should hold off on backporting this to a stable distro for a while. After all, we only updated Rolling this morning.

That seems like a good idea to me 👍

bperseghetti:
> Regardless of the outcome, I feel like we should hold off on backporting this to a stable distro for a while. After all, we only updated Rolling this morning.

I think that's reasonable.

bperseghetti:
> Here's an initial run of comprehensive CI, I did not disable linter tests, which I used to do when testing stable distros due to linter drift but I think that we've mostly moved to using Ubuntu-distributed linters so that may be less of an issue now. We'll find out.
> 
> * Linux [Build Status](http://ci.ros2.org/job/ci_linux/18459/)
> * Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13005/)
> * Windows [Build Status](http://ci.ros2.org/job/ci_windows/19152/)

Yay!!! It passed!

bperseghetti:
@clalancette @audrow @nuclearsandwich you guys think it's use on Iron has proven "stable enough" now for us to merge?

audrow:
> @audrow what I think we should do here is to merge this in, and then immediately do a new release of cyclonedds into Humble from the 0.10.x branch. Then we can let that sit in testing for a few weeks before doing a patch release. Does that make sense? Can you take care of that?

I can do that. I've been sick this week. I'll probably get to it early next week.

audrow:
Release PR opened: https://github.com/ros/rosdistro/pull/38309

Issue #1094, opened 2021-02-26
Using Python Virtual Environments


jandaa:
## Bug report

I'm trying to use a python virtual environment in my colcon workspace to separate system level and node level dependencies. I'm following the instructions given in this [tutorial](https://index.ros.org/doc/ros2/Tutorials/Using-Python-Packages/#installing-via-a-virtual-environment) in the ros2 documentation (Specifically installing via a virtual environment). However, when I check what python interpreter is being used inside the node, it returns the system level one instead of the virtual environment. This can be further verified by installing a package in the virtual environment that is not present in the system environment and try importing it from inside the node. 

I have tried doing a clean build with and without sourcing the virtual environment in the build shell. I have also tried using virtualenv as in the tutorial and also python -m venv. I also call my virtual environment "venv" as in the tutorial.

**Required Info:**

- Operating System: Ubuntu 20.04
- Installation type: Binaries
- Version or commit hash: Foxy Fitzroy
- DDS implementation: (Default) Cyclone DDS
- Client library (if applicable): N/A

#### Steps to reproduce issue
1. Follow the steps in the [tutorial](https://index.ros.org/doc/ros2/Tutorials/Using-Python-Packages/#installing-via-a-virtual-environment)
2. Create a ros2 python package with ```cd src && ros2 pkg create venv --build-type ament_python```
3. Create a python node called ```venv.py``` under src/venv/venv with the following code
```
import rclpy
from rclpy.node import Node
import sys

class ExampleNode(Node):
    def __init__(self):
        super().__init__('sensors')
        self.get_logger().info("ExampleNode started")
        self.get_logger().info(f"python path: {sys.executable}")

def main(args=None):
    rclpy.init(args=args)
    node = ExampleNode()
    rclpy.spin(node)
    rclpy.shutdown()

if __name__ == '__main__':
    main()
```
4. Add the node to setup.py with
```
entry_points={
      'console_scripts': [
          'venv = venv.venv:main'
      ],
  },
```
5. Build using
```
source ./venv/bin/activate
source /opt/ros/foxy/setup.bash
colcon build
source install/setup.sh
```
6. Run using ```ros2 run venv venv ```

#### Expected behavior
The returned python path should show ```${workspaceFolder}/venv/bin/python``` and any packages installed in the virtual environment should be accessible inside the node. 

#### Actual behavior
Returns the system base python ```/usr/bin/python3``` and breaks when trying to import a package that is installed in the virtual environment but not at system level.

[...]

PR #1524, opened 2024-02-15
Remove python_cmake_module from ros2.repos.

clalancette:
This PR aims to remove `python_cmake_module` from the core ROS 2 repositories.

The basic idea behind this is that `python_cmake_module` is an unnecessary indirection to `find_package(Python3)`, which is a standard CMake feature.  So we may as well remove it and save that additional layer of indirection.

While we are doing this, we also notice that we can support venvs in ROS 2 much better by taking advantage of some newer features of `find_package(Python3)`.  In particular, we can set [CMP0094](https://cmake.org/cmake/help/latest/policy/CMP0094.html), which sets `Python3_FIND_STRATEGY` to `LOCATION`.  In short, this means that `find_package(Python3)` will stop searching after it finds the first python3 binary it can.

However, `find_package(Python3)`, at least on Linux, will attempt to find the most specific version it can by default.  That means that if both `python3.11` and `python3` are on the PATH, it will by default choose `python3.11`.  We can also control *that* behavior by setting [Python3_FIND_UNVERSIONED_NAMES](https://cmake.org/cmake/help/latest/module/FindPython3.html#hints) to `FIRST`.

The end result of all of this is that venvs on Linux should work properly, as long as the venv is activated.  On Windows, due to different logic, this always worked properly.  If all else fails, the user can always pass `--cmake-args -DPython3_EXECUTABLE=/path/to/python3`, which will forcibly set the path to the python executable.

Given the complexity of Python packaging, I don't expect it will fix all related problems, but it should improve many of our venv issues, like https://github.com/ros2/ros2cli/issues/879 , https://github.com/ament/ament_cmake/issues/502 , https://github.com/ros2/ros2/issues/1519 , https://github.com/ros2/ros2/issues/1094 , and https://github.com/ros2/ros2/issues/1430 .

Also note that this is *not* backportable to Humble, since it requires CMake 3.20 and Humble has to support CMake versions older than that.  There we are using a simpler version of this, in https://github.com/ros2/geometry2/pull/650 and https://github.com/ament/ament_cmake/pull/507 .  This might be backportable to Iron, but we'll have to see.

Note that this is a draft because we need to get all of the dependent PRs in first, and I still have to write up the documentation and CI code for this.

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -263,10 +263,6 @@ repositories:
     type: git
     url: https://github.com/ros2/pybind11_vendor.git
     version: rolling
-  ros2/python_cmake_module:
-    type: git
-    url: https://github.com/ros2/python_cmake_module.git
-    version: rolling
   ros2/rcl:
     type: git
     url: https://github.com/ros2/rcl.git


yashi:
Just wanted to say that if you have `libpython3.12-dev` installed but your `python3` is `3.11.8`, the following two find different versions.  This is the current situation on Debian Sid right now.

- `find_package(Python3 REQUIRED COMPONENTS Development)`
- `find_package(Python3 REQUIRED COMPONENTS Interpreter)`

This leads to a build failure on some ROS related packages even with this PR applied.

https://gist.github.com/yashi/0661b04106661e75b1749d3f057331e7

yashi:
Just wanted to say that if you have `libpython3.12-dev` installed but your `python3` is `3.11.8`, the following two find different versions.  This is the current situation on Debian Sid right now.

- `find_package(Python3 REQUIRED COMPONENTS Development)`
- `find_package(Python3 REQUIRED COMPONENTS Interpreter)`

This leads to a build failure on some ROS related packages even with this PR applied.

https://gist.github.com/yashi/0661b04106661e75b1749d3f057331e7

clalancette:
> Just wanted to say that if you have `libpython3.12-dev` installed but your `python3` is `3.11.8`, the following two find different versions. This is the current situation on Debian Sid right now.

Yeah, there is just nothing we can do about that.  That is a bug in the distribution.

clalancette:
All right, I've now updated this series.  Here's is CI again on all of this, including Windows Debug:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21226/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15536/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/700/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21947/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/21948/)

clalancette:
I've updated the series again, and added in some more fixes for Windows Debug.  Here's another Windows Debug run:

* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/22062/)

clalancette:
Updated yet again, here is another CI run:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21868/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16138/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1375/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22580/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22581/)

ahcorde (approved):

PR #1570, opened 2024-05-29
Set Fast-DDS version to 2.14.x

MiguelCompany:
With https://github.com/ros2/rmw_fastrtps/pull/756 in, it is again safe to use 2.14.x branch of Fast DDS.

Backport to Jazzy would be possible since the fix was also backported in https://github.com/ros2/rmw_fastrtps/pull/759

PR #1574, opened 2024-06-25
Update Fast-RTPS to 2.14.1.

clalancette:
That should match what we are going to release into Rolling binary packages.

### DIFF
--- a/ros2.repos
+++ b/ros2.repos
@@ -34,7 +34,7 @@ repositories:
   eProsima/Fast-DDS:
     type: git
     url: https://github.com/eProsima/Fast-DDS.git
-    version: c2a4523a21a94302298bc995057719546b72df2d
+    version: v2.14.2
   eProsima/foonathan_memory_vendor:
     type: git
     url: https://github.com/eProsima/foonathan_memory_vendor.git
--- a/ros2.repos
+++ b/ros2.repos
@@ -34,7 +34,7 @@ repositories:
   eProsima/Fast-DDS:
     type: git
     url: https://github.com/eProsima/Fast-DDS.git
-    version: v2.14.2
+    version: v2.14.1
   eProsima/foonathan_memory_vendor:
     type: git
     url: https://github.com/eProsima/foonathan_memory_vendor.git


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21321/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15634/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/827/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22050/)

ros2.repos
```diff
@@ -34,7 +34,7 @@ repositories:
   eProsima/Fast-DDS:
     type: git
     url: https://github.com/eProsima/Fast-DDS.git
-    version: c2a4523a21a94302298bc995057719546b72df2d
+    version: v2.14.2
```
fujitatomoya, line 37: No such branch???

```suggestion
    version: 2.14.2
```

fujitatomoya (changes requested):

fujitatomoya:
@clalancette i think we should take https://github.com/ros2/ros2/pull/1570 to rolling instead?

CC: @MiguelCompany

ros2.repos
clalancette: > No such branch???

This isn't a branch, it's a tag: https://github.com/eProsima/Fast-DDS/tree/v2.14.2

clalancette:
> @clalancette i think we should take #1570 to rolling instead?

Well, that one is failing CI.  But the version we currently have released into Rolling (2.14.0) doesn't actually work at all.  So I'm hoping to make the version released into CI (via https://github.com/ros/rosdistro/pull/41796) match what is in ros2.repos, and have them both work.

clalancette:
Oof, this version is failing CI, much like #1570.  Let's see if v2.14.1 is any better.

clalancette:
Using 2.14.1:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21322/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15635/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/828/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22051/)

clalancette:
All right, it looks like 2.14.1 is much better here.  I've updated the title, this needs review.  I'll also close out https://github.com/ros/rosdistro/pull/41796, and instead release 2.14.1 there.

christophebedard (approved):
The RHEL test failure isn't new. I can't find the same Windows test failure in any recent (nightly) CI, but it seems similar to some recent failures.

clalancette:
> The RHEL test failure isn't new. I can't find the same Windows test failure in any recent (nightly) CI, but it seems similar to some recent failures.

Thanks for checking!  Yes, you are correct that both of those failures are unrelated here, and have been seen on the CI farm before.  With that, I'm going to go ahead and merge this one in, thanks for the review.

PR #1642, opened 2025-02-03
Add a pixi.toml file for installation on Windows.

clalancette:

Note that the package versions in here are exhaustively specified, and specified to be as close to their counterparts in Ubuntu as possible.  Where this is not possible, we choose the next closest version that was experimentally determined to work.  Where packages don't exist, they are generally covered via vendor packages in ros2.repos.

Note that there will be a separate pixi.toml file per supported ROS distribution, and thus hence per-branch.  This PR should *not* be directly backported; each distribution requires a separate PR, which I'll open once this one has been approved.

Finally, this PR *must* go in before either of the counterparts, since they both depend on this file existing.  Conversely, this PR can be merged at any time since it won't actually cause changes to happen until both of those other PRs are merged.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -109,7 +109,7 @@ spdlog = "==1.12.0"
 sqlite = "==3.45.2"  # TODO: Conda has 3.45.1, but it is not installable with Python 3.12
 tinyxml2 = "==10.0.0"
 typing_extensions = "==4.10.0"
-# TODO: Conda is missing uncrustify==0.78.1
+uncrustify = "==0.78.1"
 vcstool = "==0.3.0"
 yaml-cpp = "==0.8.0"
 yamllint = "==1.33.0"
--- a/pixi.toml
+++ b/pixi.toml
@@ -1,0 +1,112 @@
+[project]
+name = "pixi_ros2_rolling"
+version = "0.1.0"
+description = "Dependencies to build ROS 2 on Windows"
+authors = ["Chris Lalancette <clalancette@gmail.com>"]
+channels = ["conda-forge"]
+platforms = ["win-64"]
+
+[target.win-64.activation]
+env = {QT_QPA_PLATFORM_PLUGIN_PATH="%cd%\\.pixi\\envs\\default\\Library\\plugins\\platforms"}
+
+[dependencies]
+7zip = ">=23.1,<24"
+argcomplete = "==3.1.4"
+asio = "==1.28.1"
+assimp = "==5.3.1"
+benchmark = "==1.8.3"
+bullet = "==3.25"  # TODO: Conda has 3.24, but it is not installable with Python 3.12
+catkin_pkg = "==1.0.0"
+cmake = "==3.28.3"
+colcon-cmake = ">=0.2.28,<0.3"
+colcon-core = ">=0.17.1,<0.18"
+colcon-defaults = ">=0.2.8,<0.3"
+colcon-library-path = ">=0.2.1,<0.3"
+colcon-metadata = ">=0.2.5,<0.3"
+colcon-mixin = ">=0.2.3,<0.3"
+colcon-output = ">=0.2.13,<0.3"
+colcon-package-information = ">=0.4.0,<0.5"
+colcon-package-selection = ">=0.2.10,<0.3"
+colcon-parallel-executor = ">=0.2.4,<0.3"
+colcon-pkg-config = ">=0.1.0,<0.2"
+colcon-powershell = ">=0.4.0,<0.5"
+colcon-python-setup-py = ">=0.2.7,<0.3"
+colcon-recursive-crawl = ">=0.2.3,<0.3"
+colcon-ros = ">=0.5.0,<0.6"
+colcon-ros-domain-id-coordinator = ">=0.2.1,<0.3"
+colcon-test-result = ">=0.3.8,<0.4"
+console_bridge = "==1.0.1"
+coverage = "==7.4.4"
+cppcheck = "==2.15.0"  # TODO: This version doesn't actually work, so we should just remove ament_cppcheck
+cryptography = "==41.0.7"
+cunit = "==2.1.3"
+curl = "==8.5.0"
+distlib = "==0.3.8"
+docutils = "==0.20.1"
+eigen = "==3.4.0"
+empy = "==3.3.4"
+flake8 = "==7.0.0"
+flake8-blind-except = "==0.2.1"
+flake8-builtins = "==2.1.0"
+flake8-class-newline = "==1.6.0"
+flake8-comprehensions = "==3.14.0"
+flake8-deprecated = "==2.2.1"
+flake8-docstrings = "==1.6.0"
+flake8-import-order = "==0.18.2"
+flake8-quotes = "==3.4.0"
+git = "==2.43.0"
+graphviz = "==9.0.0"  # TODO: pygraphviz 1.11 needs at least graphviz 9.0.0
+importlib-metadata = "==8.5.0"  # TODO: Conda doesn't have 4.12.0
+iniconfig = "==1.1.1"
+lark = "==1.1.9"
+libcurl = "==8.5.0"
+lxml = "==5.2.1"
+lz4-c = "==1.9.4"
+mccabe = "==0.7.0"
+mypy = "==1.9.0"
+mypy_extensions = "==1.0.0"
+numpy = "==1.26.4"
+opencv = "==4.9.0"  # TODO: Conda has 4.6.0, but it is not installable with Python 3.12
+openssl = "==3.3.2"  # TODO: Conda doesn't have 3.0.13
+orocos-kdl = "==1.5.1"
+packaging = "==24.0"
+pathspec = "==0.12.1"
+pip = "==24.0"
+pluggy = "==1.4.0"
+psutil = "==5.9.8"
+pybind11 = "==2.11.1"
+pycodestyle = "==2.11.1"
+pydocstyle = "==6.3.0"
+pydot = "==1.4.2"
+pyflakes = "==3.2.0"
+pygraphviz = "==1.11"  # TODO: conda doesn't have pygraphviz 1.7 built for Python 3.12
+pyparsing = "==3.1.1"
+pyqt = "==5.15.9"  # TODO: Conda doesn't have 5.15.10
+pyqt5-sip = "==12.12.2"  # TODO: Conda doesn't have pyqt5-sip 12.13.0
+pytest = "==7.4.4"
+pytest-cov = "==4.1.0"
+pytest-mock = "==3.12.0"
+pytest-repeat = "==0.9.2"  # TODO: Conda doesn't have 0.9.3
+pytest-rerunfailures = "==12.0"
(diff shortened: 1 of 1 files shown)


codebot (approved):

pixi.toml
```diff
@@ -0,0 +1,112 @@
...
+version = "0.1.0"
+description = "Dependencies to build ROS 2 on Windows"
+authors = ["Chris Lalancette <clalancette@gmail.com>"]
+channels = ["conda-forge"]
+platforms = ["win-64"]
+
+[target.win-64.activation]
+env = {QT_QPA_PLATFORM_PLUGIN_PATH="%cd%\\.pixi\\envs\\default\\Library\\plugins\\platforms"}
+
+[dependencies]
```
nuclearsandwich, line 12: > Note that the package versions in here are exhaustively specified, and specified to be as close to their counterparts in Ubuntu as possible. Where this is not possible, we choose the next closest version that was experimentally determined to work. Where packages don't exist, they are generally covered via vendor packages in ros2.repos.

What rationale was used for the packages that have semver restrictions instead of specific pins e.g. 7zip and colcon libs but not catkin_pkg?

nuclearsandwich (approved):

pixi.toml
clalancette: That is a good question.

Basically, it is a difference between tools that we use to setup/initiate the build (7zip and colcon fall into this category), and dependencies of the ROS packages (everything else).

To put it another way, we call both 7zip and colcon from the Dockerfile (the former to extract Connext, the latter to actually run the build).  Since these are the build tooling we assume that they will continue to do the same job, so it is safe to let them upgrade.  It is doubly-safe in the colcon case, since we are in charge of its development.

All other packages here are dependencies of one or more ROS packages.  As such, we aren't sure that the API/ABI will be stable across their versions, so we pin them more aggressively.

Certainly we could consider other strategies here, but that is the one I chose for now.  Does that make sense to you?  Do you think we should consider a different strategy?

nuclearsandwich: This makes sense as a strategy. I think it would be worth documenting, either by separating environmental dependencies from build / build-tooling dependencies via sort order or individual line comments. If e.g. colcon-core becomes a runtime dependencies of ros2cli or 7zip becomes a buildtool depend of a vendor package their status would change and it wouldn't be clear that the version constraint should be changed as a result.

pixi.toml
```diff
@@ -0,0 +1,112 @@
...
+pydot = "==1.4.2"
+pyflakes = "==3.2.0"
+pygraphviz = "==1.11"  # TODO: conda doesn't have pygraphviz 1.7 built for Python 3.12
+pyparsing = "==3.1.1"
+pyqt = "==5.15.9"  # TODO: Conda doesn't have 5.15.10
+pyqt5-sip = "==12.12.2"  # TODO: Conda doesn't have pyqt5-sip 12.13.0
+pytest = "==7.4.4"
+pytest-cov = "==4.1.0"
+pytest-mock = "==3.12.0"
+pytest-repeat = "==0.9.2"  # TODO: Conda doesn't have 0.9.3
```
j-rivero, line 89: We should now have this one https://github.com/conda-forge/pytest-repeat-feedstock/pull/4

pixi.toml
```diff
@@ -0,0 +1,112 @@
...
+pyyaml = "==6.0.1"
+qt = "==5.15.8"  # TODO: Conda doesn't have 5.15.13
+setuptools = "==68.1.2"
+six = "==1.16.0"
+snowballstemmer = "==2.2.0"
+spdlog = "==1.12.0"
+sqlite = "==3.45.2"  # TODO: Conda doesn't have 3.45.1
+tinyxml2 = "==10.0.0"
+typing_extensions = "==4.10.0"
+# TODO: Conda is missing uncrustify==0.78.1
```
j-rivero, line 106: blocked by https://github.com/conda-forge/uncrustify-feedstock/pull/14

pixi.toml
```diff
@@ -0,0 +1,112 @@
...
+flake8-builtins = "==2.1.0"
+flake8-class-newline = "==1.6.0"
+flake8-comprehensions = "==3.14.0"
+flake8-deprecated = "==2.2.1"
+flake8-docstrings = "==1.6.0"
+flake8-import-order = "==0.18.2"
+flake8-quotes = "==3.4.0"
+git = "==2.43.0"
+graphviz = "==9.0.0"  # TODO: pygraphviz 1.11 needs at least graphviz 9.0.0
+importlib-metadata = "==8.5.0"  # TODO: Conda doesn't have 4.12.0
```
j-rivero, line 59: 4.12.0 does not exists but the closest would be 4.13.0. I was able to build this environment with 4.13.0, not sure if there is any runtime incompatibility though.

pixi.toml
```diff
@@ -0,0 +1,112 @@
...
+lark = "==1.1.9"
+libcurl = "==8.5.0"
+lxml = "==5.2.1"
+lz4-c = "==1.9.4"
+mccabe = "==0.7.0"
+mypy = "==1.9.0"
+mypy_extensions = "==1.0.0"
+numpy = "==1.26.4"
+opencv = "==4.9.0"  # TODO: Conda has 4.6.0, but it is not installable with Python 3.12
+openssl = "==3.3.2"  # TODO: Conda doesn't have 3.0.13
```
j-rivero, line 70: I think that conda-forge has the 3.0.13 but it is incompatible with python 3.12.3. 
```  × failed to solve the conda requirements of 'default' 'win-64'
  ╰─▶ Cannot solve the request because of: The following packages are incompatible
      ├─ openssl ==3.0.13 can be installed with any of the following options:
      │  └─ openssl 3.0.13 | 3.0.13
      └─ python ==3.12.3 cannot be installed because there are no viable options:
         └─ python 3.12.3 | 3.12.3 would require
            └─ openssl >=3.2.1,<4.0a0, which cannot be installed because there are no viable options:
               └─ openssl 3.2.1 | 3.2.1 | 3.2.1 | 3.2.2 | 3.2.3 | 3.3.0 | 3.3.0 | 3.3.0 | 3.3.0 | 3.3.1 | 3.3.1 | 3.3.1 | 3.3.1 | 3.3.2 | 3.3.2 | 3.4.0 |
      3.4.0, which conflicts with the versions reported above.
```

pixi.toml
```diff
@@ -0,0 +1,112 @@
...
+pygraphviz = "==1.11"  # TODO: conda doesn't have pygraphviz 1.7 built for Python 3.12
+pyparsing = "==3.1.1"
+pyqt = "==5.15.9"  # TODO: Conda doesn't have 5.15.10
+pyqt5-sip = "==12.12.2"  # TODO: Conda doesn't have pyqt5-sip 12.13.0
+pytest = "==7.4.4"
+pytest-cov = "==4.1.0"
+pytest-mock = "==3.12.0"
+pytest-repeat = "==0.9.2"  # TODO: Conda doesn't have 0.9.3
+pytest-rerunfailures = "==12.0"
+pytest-runner = "==6.0.0"  # TODO: Conda doesn't have 2.11.1
```
j-rivero, line 91:
```suggestion
pytest-runner = "==6.0.0"  # TODO: Conda doesn't have 2.11.1 built for Python 3.12
```

pixi.toml
```diff
@@ -0,0 +1,112 @@
...
+python-dateutil = "==2.8.2"
+python-fastjsonschema = "==2.19.0"
+python-orocos-kdl = "==1.5.1"
+pyyaml = "==6.0.1"
+qt = "==5.15.8"  # TODO: Conda doesn't have 5.15.13
+setuptools = "==68.1.2"
+six = "==1.16.0"
+snowballstemmer = "==2.2.0"
+spdlog = "==1.12.0"
+sqlite = "==3.45.2"  # TODO: Conda doesn't have 3.45.1
```
j-rivero, line 103:
```suggestion
sqlite = "==3.45.2"  # TODO: Conda doesn't have 3.45.1 built for Python 3.12
```

j-rivero (commented):
give a quick pass over the missing pieces, looks good with some suggestions

pixi.toml
clalancette: That PR was merged, and it looked like things were built.  Unfortunately, that build uncrustify 0.77, and uncrustify is very sensitive to the version.  So I think we are going to have to continue to use our vendored package until uncrustify 0.78.1 is available.

pixi.toml
```diff
@@ -0,0 +1,112 @@
+[project]
+name = "pixi_ros2_rolling"
+version = "0.1.0"
+description = "Dependencies to build ROS 2 on Windows"
+authors = ["Chris Lalancette <clalancette@gmail.com>"]
+channels = ["conda-forge"]
+platforms = ["win-64"]
+
+[target.win-64.activation]
+env = {QT_QPA_PLATFORM_PLUGIN_PATH="%cd%\\.pixi\\envs\\default\\Library\\plugins\\platforms"}
```
hameerabbasi, line 10:
```suggestion
[target.win-64.activation.env]
QT_QPA_PLATFORM_PLUGIN_PATH = "%cd%\\.pixi\\envs\\default\\Library\\plugins\\platforms"
```

pixi.toml
clalancette: Updated in f15a86444626977cbc8ad4eff9c6ccb7e3de65d0

pixi.toml
clalancette: Updated in f15a86444626977cbc8ad4eff9c6ccb7e3de65d0

pixi.toml
clalancette: Updated the comment in f15a86444626977cbc8ad4eff9c6ccb7e3de65d0 .

pixi.toml
clalancette: Good call.  Updated in f15a86444626977cbc8ad4eff9c6ccb7e3de65d0.

pixi.toml
clalancette: Ah, cool, thanks.  I'll keep an eye on that one and give it a shot.

clalancette: And it looks like it is installable now.  Added to the list in 0f96d37b42253f3a93ecab172279f01ead6e135e , thanks!

pixi.toml
clalancette: In f15a86444626977cbc8ad4eff9c6ccb7e3de65d0, I ended up splitting them into two groups and adding in comments explaining the strategy in each group.

pixi.toml
clalancette: Changed in fa920d8e2414daa3d20f6bccc212c9dc7f52d98c .

clalancette:
All right, I've made all of the changes requested here, and retested everything locally.  Things seem to be happy.

@j-rivero If you are happy with the current state of this PR, I'd appreciate an approval.  After that I'll go ahead and merge this, then open up the equivalent for Jazzy and Humble.

j-rivero:
> @j-rivero If you are happy with the current state of this PR, I'd appreciate an approval. After that I'll go ahead and merge this, then open up the equivalent for Jazzy and Humble.

Thanks for the fixes Chris ! Good to go !

ros-discourse:

PR #1643, opened 2025-02-04
[Jazzy]: Add a pixi.toml file for installation on Windows.

clalancette:

This is the Jazzy version of #1642 .  Compared to the Rolling pixi.toml, this file is *exactly* the same, with small difference of the name.  This is on purpose, since Jazzy is very close to Rolling right now (but won't always be).

Note that the package versions in here are exhaustively specified, and specified to be as close to their counterparts in Ubuntu as possible. Where this is not possible, we choose the next closest version that was experimentally determined to work. Where packages don't exist, they are generally covered via vendor packages in ros2.repos.

Finally, this PR must go in before either of the counterparts, since they both depend on this file existing. Conversely, this PR can be merged at any time since it won't actually cause changes to happen until both of those other PRs are merged.

@j-rivero @nuclearsandwich @marcoag FYI

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -1,0 +1,118 @@
+[project]
+name = "pixi_ros2_jazzy"
+version = "0.1.0"
+description = "Dependencies to build ROS 2 on Windows"
+authors = ["Chris Lalancette <clalancette@gmail.com>"]
+channels = ["conda-forge"]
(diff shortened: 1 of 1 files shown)


j-rivero (approved):

ros-discourse:

PR #1644, opened 2025-02-05
[Humble]: Add a pixi.toml file for installation on Windows.

clalancette:

This is the Humble version of https://github.com/ros2/ros2/pull/1642 . Compared to the Rolling pixi.toml, this file has a bunch of different dependencies, making this closer to the same package versions as Ubuntu 22.04.

Note that the package versions in here are exhaustively specified, and specified to be as close to their counterparts in Ubuntu as possible. Where this is not possible, we choose the next closest version that was experimentally determined to work. Where packages don't exist, they are generally covered via vendor packages in ros2.repos.

Finally, this PR must go in before either of the counterparts, since they both depend on this file existing. Conversely, this PR can be merged at any time since it won't actually cause changes to happen until both of those other PRs are merged.

@j-rivero @nuclearsandwich @audrow FYI

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -1,0 +1,124 @@
+[project]
+name = "pixi_ros2_humble"
+version = "0.1.0"
+description = "Dependencies to build ROS 2 on Windows"
+authors = ["Chris Lalancette <clalancette@gmail.com>"]
+channels = ["conda-forge"]
+platforms = ["win-64"]
+
+[target.win-64.activation.env]
+PYTHONPATH="%cd%\\.pixi\\envs\\default\\Lib"
+QT_QPA_PLATFORM_PLUGIN_PATH="%cd%\\.pixi\\envs\\default\\Library\\plugins\\platforms"
+
+[dependencies]
+# The following are "tooling" dependencies, used to install or initiate builds.
+# We pin them less aggressively since they will presumably continue to their job,
+# and we don't rely on any particular API or ABI for them to be stable.
+7zip = ">=23.1,<24"
+colcon-cmake = ">=0.2.28,<0.3"
+colcon-core = ">=0.17.1,<0.18"
+colcon-defaults = ">=0.2.8,<0.3"
+colcon-library-path = ">=0.2.1,<0.3"
+colcon-metadata = ">=0.2.5,<0.3"
+colcon-mixin = ">=0.2.3,<0.3"
+colcon-output = ">=0.2.13,<0.3"
+colcon-package-information = ">=0.4.0,<0.5"
+colcon-package-selection = ">=0.2.10,<0.3"
+colcon-parallel-executor = ">=0.2.4,<0.3"
+colcon-pkg-config = ">=0.1.0,<0.2"
+colcon-powershell = ">=0.4.0,<0.5"
+colcon-python-setup-py = ">=0.2.7,<0.3"
+colcon-recursive-crawl = ">=0.2.3,<0.3"
+colcon-ros = ">=0.5.0,<0.6"
+colcon-ros-domain-id-coordinator = ">=0.2.1,<0.3"
+colcon-test-result = ">=0.3.8,<0.4"
+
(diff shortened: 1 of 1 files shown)


j-rivero (approved):
The most notable change with respect to jammy can be the openssl version. I'm afraid that for bumping it to something > 3.0.0 the amount of changes in other important pinned versions (like qt5 or opencv) is too much so I agree that this set is probably one of the best approaches. I left minor comments with respect about how we can capture the inter-blocking, @clalancette handle them as you prfer.

pixi.toml
```diff
@@ -0,0 +1,124 @@
...
+colcon-ros-domain-id-coordinator = ">=0.2.1,<0.3"
+colcon-test-result = ">=0.3.8,<0.4"
+
+# The rest of the dependencies here are used by one or more ROS packages.  We aggressively
+# pin them to particular versions that are as close as possible to their counterparts in Ubuntu.
+argcomplete = "==1.9.4"  # TODO: conda doesn't have argcomplete 1.8.1
+asio = "==1.22.1"  # TODO: conda doesn't have asio 1.18.1
+assimp = "==5.2.3"  # TODO: conda doesn't have assimp 5.2.2
+benchmark = "==1.6.1"
+bullet = "==3.21"  # TODO: conda doesn't have bullet 3.06
```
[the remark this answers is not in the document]
clalancette: Fixed in 0f4bcdd27d89e89e9c4f06df5f6ed35705da3c0b

pixi.toml
```diff
@@ -0,0 +1,124 @@
...
+libcurl = "==7.81.0"
+lxml = "==4.9.2"  # TODO: lxml 4.8.0, 4.9.0, and 4.9.1 crash during import with conda
+lz4-c = "==1.9.3"
+mccabe = "==0.6.1"
+mypy = "==0.942"
+mypy_extensions = "==0.4.3"
+netifaces = "==0.11.0"
+numpy = "==1.21.5"
+opencv = "==4.5.5"  # TODO: conda doesn't have opencv 4.5.4
+openssl = "==1.1.1w"  # TODO: conda doesn't have openssl 3.0.2
```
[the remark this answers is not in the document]
clalancette: I added a more detailed comment in 0f4bcdd27d89e89e9c4f06df5f6ed35705da3c0b .

clalancette:
Since the only changes since review here are comments, I'm going ahead and merging this.  Thanks @j-rivero !

ros-discourse:

pixi.toml
```diff
@@ -0,0 +1,124 @@
...
+console_bridge = "==1.0.1"
+coverage = "==6.2"
+cppcheck = "==2.10.1"  # TODO: This version doesn't actually work, so we should just remove ament_cppcheck
+cryptography = "==3.4.8"
+cunit = "==2.1.3"  # TODO: conda doesn't have cunit 2.1
+curl = "==7.81.0"
+distlib = "==0.3.4"
+docutils = "==0.17.1"
+eigen = "==3.4.0"
+empy = "==3.3.4"
```
christophebedard, line 54: REP 2000 for Windows @ Humble sets EmPY to 3.3.2: https://www.ros.org/reps/rep-2000.html#humble-hawksbill-may-2022-may-2027. Should we update it to 3.3.4 in REP 2000?

clalancette: > REP 2000 for Windows (all non-Ubuntu platforms) @ Humble sets EmPY to 3.3.2: https://www.ros.org/reps/rep-2000.html#humble-hawksbill-may-2022-may-2027. Should we update it to 3.3.4 in REP 2000?

Good call.  Yes, we should update it (and all of the rest of the dependencies) in REP-2000 now, for Humble, Jazzy, and Kilted.

christophebedard: Thanks for confirming. I've opened https://github.com/ros-infrastructure/rep/issues/416

PR #1771, opened 2025-12-03
Ignore generated pixi directory

mjcarroll:
When working in this repository, prevent inadvertently checking in the `.pixi` directory, which will contain the installed environments.


fujitatomoya (approved):
fix looks good to me.

PR #1772, opened 2025-12-03
Update gitattributes to pixi best practice

mjcarroll:
This is in preparation of committing pixi.lock files, which shouldn't be diff'd with normal tooling.


.gitattributes
```diff
@@ -0,0 +1,2 @@
+# SCM syntax highlighting & preventing 3-way merges
+pixi.lock merge=binary linguist-language=YAML linguist-generated=true
```
fujitatomoya: Add an empty line (newline character) at the end of the file?

ahcorde:
```suggestion
pixi.lock merge=binary linguist-language=YAML linguist-generated=true

```

ahcorde (approved):

PR #1773, opened 2025-12-03
draft: unpin windows dependencies

mjcarroll:
Some other quality of life stuff as well

To get started here:

Get `buildcache`:

https://gitlab.com/bits-n-bites/buildcache/-/releases

And put it in C:\buildcache (or somewhere else and adjust paths, hardcoded for now)

```
# Install both runtime and devel dependencies
pixi install -a

# Checkout source repos
pixi run import

# use a pixi task to build
pixi run build --packages-up-to rclcpp

# OR Enter the pixi shell with devel dependencies
pixi shell -e dev
```

### DIFF
--- a/.github/workflows/pixi_ci.yaml
+++ b/.github/workflows/pixi_ci.yaml
@@ -14,13 +14,11 @@ jobs:
           environments: dev
           pixi-version: v0.60.0
           cache: true
-          cache-key: ${{ runner.os }}-${{ hashFiles('pixi.lock') }}
+          cache-key: ${{ matrix.os }}-${{ hashFiles('pixi.lock') }}
           frozen: true
-      - uses: mjcarroll/buildcache-action@v2
+      - uses: hendrikmuhs/ccache-action@v1.2
         with:
-          cache_key: ${{ matrix.label }}
-          upload_buildcache_log: 'false'
-          zero_buildcache_stats: 'true'
+          key: ${{ matrix.os }}
       - name: Install sources
         run: pixi run import
       - name: Build something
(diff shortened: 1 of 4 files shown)


traversaro:
Just FYI, at my org we obtained good result in caching compilation steps with pixi+colcon with ccache, see https://github.com/robotology/robotology-superbuild/pull/1920 (that is not colcon but we also did that in internal colcon projects, fyi @xela-95 ).

ros2.repos
```diff
@@ -66,7 +66,7 @@ repositories:
   ros-perception/image_common:
```
ahcorde: restore this file

mjcarroll:
> Just FYI, at my org we obtained good result in caching compilation steps with pixi+colcon with ccache, see [robotology/robotology-superbuild#1920](https://github.com/robotology/robotology-superbuild/pull/1920) (that is not colcon but we also did that in internal colcon projects, fyi @xela-95 ).

Did you find that ccache worked on Windows as well?  That's the reason that I have been using buildcache in the past, but would be willing to switch if it "just works (TM)".

traversaro:
> > Just FYI, at my org we obtained good result in caching compilation steps with pixi+colcon with ccache, see [robotology/robotology-superbuild#1920](https://github.com/robotology/robotology-superbuild/pull/1920) (that is not colcon but we also did that in internal colcon projects, fyi @xela-95 ).
> 
> Did you find that ccache worked on Windows as well? That's the reason that I have been using buildcache in the past, but would be willing to switch if it "just works (TM)".

Yes, it "worked out of the box" also on Windows for us (as long as you do not use clang-cl, but just regular cl). The modification required should be just add under the `[activation.env]` table:

~~~
CMAKE_C_COMPILER_LAUNCHER="ccache"
CMAKE_CXX_COMPILER_LAUNCHER="ccache"
CCACHE_BASEDIR="$PIXI_PROJECT_ROOT"
~~~

and then add `ccache = "*"` as dependency.

mjcarroll:
Ah, that's excellent, because buildcache isn't available as a package, making some of the logic here kind of weird.  I will give this a shot.

PR #1774, opened 2025-12-04
Add freetype dependency version 2.13.3

ahcorde:
Add freetype dependency version 2.13.3

Related with https://github.com/ros2/rviz/pull/1636/files

To be able to merge the PR in https://github.com/ros2/rviz/pull/1636/files we should include freetype


mjcarroll:
I think we also need `zlib` for ogre to build?

ahcorde:
> I think we also need `zlib` for ogre to build?

I build this locally and it's working, maybe zlib is installed with another package ?

mjcarroll (approved):

PR #1810, opened 2026-03-31
Pixi: Removed Qt5 and add Qt6 support

ahcorde:
Pixi: Removed Qt5 and add Qt6 support


ahcorde:
Related with https://github.com/ros2/rviz/pull/1693

mjcarroll (approved):

PR #1814, opened 2026-04-08
Kilted: Pinned FastCDR version

ahcorde:
Kilted: Pinned FastCDR version. 

This change https://github.com/eProsima/Fast-CDR/pull/303 only make sense on Rolling not Kilted


ahcorde:
Superseed by https://github.com/ros2/ros2/pull/1815

PR #1822, opened 2026-04-30
Use vcs2l when validating repos files

cottsay:
The old `vcstool` has a bug that causes `validate` to fail for SHA refs. The bug appears to be fixed in `vcs2l`, so I think it's time to move forward.


christophebedard (approved):

cottsay:
@Mergifyio backport lyrical lyrical-release kilted kilted-release jazzy jazzy-release humble humble-release

PR #1854, opened 2026-08-21
Add ninja and sccache in a buildfarm-only environment

mjcarroll:
The buildfarm is moving its Windows jobs onto Ninja with sccache as a compiler launcher (ros2/ci#900, ros2/ci#899), and gets its tooling from this file. Neither tool is needed to build ROS 2, so neither belongs in `[dependencies]` — that installs them for every developer who uses this manifest. They go in a feature instead, with an environment asked for by name:

```console
pixi install -e buildfarm
pixi run -e buildfarm ...
```

An environment is the default feature plus the features it names, so `buildfarm` is everything in `[dependencies]` and these two on top. The default environment is unchanged: the regenerated lock has 281 packages for `default`, none of them ninja or sccache, and 284 for `buildfarm`.

Ninja is not optional for the cache, which is why both arrive together: CMake honours `CMAKE_<LANG>_COMPILER_LAUNCHER` only under Ninja and Makefiles, and silently ignores it under Visual Studio.

The first commit fixes a latent bug this exposed. `QT_QPA_PLATFORM_PLUGIN_PATH` was spelled out as `.pixi/envs/default/...`, correct only for the default environment; any other one would be pointed at a different package set and `qt_gui_cpp` would fail to find a platform plugin at runtime. `%CONDA_PREFIX%` is whichever environment pixi activated. Note the percent form — pixi does not expand `$CONDA_PREFIX` in `activation.env`, it passes it through literally. Both tested against pixi 0.78.0.

Both packages are pinned loosely, as tooling whose APIs we do not depend on.


cottsay:
Is there some way we can make this only install stuff on the buildfarm? Every ROS user that uses this pixi.toml is going to get these.

mjcarroll:
Yeah, I think we could shove it in an environment, let me take a look.

mjcarroll:
Okay, it's now using an environment, so you have to specifically opt into installing it (using `-e buildfarm`) on the command line.  There will have to be a corresponding ros2/ci PR to actually make use of it (though that was already needed to change the generator).

dskkato:
Could we consider including Ninja in the default pixi environment as well? On Windows, it is also useful for generating `compile_commands.json`, since CMake's Visual Studio generators do not support `CMAKE_EXPORT_COMPILE_COMMANDS`. This file is needed to ues ament_clang_tidy on Windows.

dskkato:
Just to add a quick note: after looking into this a bit more, it seems Ninja is not strictly required here for Windows via pixi. At least in my local Windows environment and on GitHub Actions, I was able to use Ninja without installing it separately.

So I think my earlier assumption that it needed to be included explicitly in the default pixi environment was a bit too strong. Sorry.

clalancette (approved):

PR #1858, opened 2026-08-29
Validate pixi.toml solves and pixi.lock matches, on every pull request

mjcarroll:
A change to `pixi.toml` invalidates the `pixi.lock` next to it, and nothing notices. This adds a check that refuses the merge unless both are satisfied: the manifest solves on every platform it declares, and the lock file in the pull request is the one that solve produces.

It also adds the lock file itself, which the repository has never carried.

### Add a pixi.lock

`pixi.toml` has been here since the pixi environment landed, but the lock it produces never was, so nothing recorded which packages a given manifest actually resolved to — two people running `pixi install` a week apart could get different environments.

Generated with pixi 0.60.0, covering both platforms the manifest declares. I confirmed it is reproducible: solving the same manifest in an unrelated directory produced a byte-identical file. `.gitattributes` already marks `pixi.lock` as `merge=binary` and `linguist-generated`, so it stays collapsed in diffs.

### Check it on every pull request

`pixi lock --check` does both halves of the check in one command. It solves each platform under `workspace.platforms` and exits non-zero when a platform has no solution *or* when the result differs from the committed lock.

It also rewrites `pixi.lock` as it goes, which is what lets the workflow tell those two failures apart — an unsolvable manifest leaves the working tree untouched, a merely stale lock does not. So a contributor gets "your manifest does not solve on win-64" rather than one opaque error. When the lock is just stale, the corrected file is uploaded as a `pixi-lock-suggested` artifact, so committing it does not require installing pixi first.

Verified behaviour, against this manifest:

| scenario | result |
| --- | --- |
| lock matches manifest | exit 0 |
| manifest gains a dependency, lock unchanged | exit 1, classified `stale` |
| a platform has no solution | exit 1, classified `unsolvable` |
| `pixi install --locked` on a stale lock | exit 1, `lock-file not up-to-date` |

### Notes on a few choices

**The install matrix is derived from `workspace.platforms`, not hard-coded.** Adding a platform to the manifest starts testing it here instead of quietly going untested, and a platform with no runner mapping fails loudly rather than being skipped. `pixi install --locked` refuses to relax the lock, so each leg installs exactly what the pull request proposes to merge — proving the solution is usable, not merely satisfiable on paper.

**Nothing is written back to the branch.** An earlier draft pushed the regenerated lock onto the pull request, but doing that for a fork needs credentials this workflow would then be holding while handling contributor-controlled input. The lock file stays the author's to commit.

**There is deliberately no `paths:` filter.** A required check that is filtered out never reports, leaving unrelated pull requests waiting on a status that never arrives. The workflow always runs and works out for itself whether there is anything to do; on a `ros2.repos`-only change it passes immediately and nothing expensive runs.

**`pixi-version` is pinned.** A lock file is only byte-identical to one another pixi produced if the two are the same version, so an unpinned runner could reject a perfectly good lock. Bump the pin and regenerate the lock together. Happy to move this to a newer pixi before merge if you'd prefer to start current.

### After merge

Mark **`Pixi Lock / Pixi lock is valid`** as a required status check — it is the aggregating job, and the only one that always reports.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/pixi-lock.yaml
+++ b/.github/workflows/pixi-lock.yaml
@@ -1,0 +1,227 @@
+---
+name: Pixi Lock
+
+# Checks that pixi.toml still solves on every platform it declares, and that
+# the pixi.lock committed alongside it is the lock file that solve produces.
+# Nothing is written back: a pull request that changes pixi.toml has to carry
+# the matching pixi.lock, and this workflow is what refuses to let it merge
+# otherwise.
+#
+# There is deliberately no `paths:` filter.  A required check that is filtered
+# out never reports, which leaves unrelated pull requests waiting on a status
+# that will never arrive, so the workflow always runs and decides internally
+# whether there is any work to do.
+
+on:
+  pull_request:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: pixi-lock-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  changes:
+    name: Detect pixi changes
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+    outputs:
+      pixi: ${{ steps.changes.outputs.pixi }}
+    steps:
+      - name: Look for pixi files in the pull request
+        id: changes
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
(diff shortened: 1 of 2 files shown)


mjcarroll:
Closing in favour of splitting this into three focused pull requests: declaring the pixi version in `pixi.toml`, adding `pixi.lock`, and adding the CI check that validates the two against each other. I will link them here as they go up.

PR #1859, opened 2026-08-29
Declare which pixi this workspace expects

mjcarroll:
A lock file's contents depend on which pixi wrote it. Before shipping one, the workspace should say which pixi it expects, or a contributor on an older release regenerates the lock, gets a legitimately different file, and has no way to tell that the version is the reason.

`>=0.78` is the current release. It is a floor, so it does not stop anyone moving forward.


christophebedard (approved):
Makes sense

christophebedard:
> It is a floor, so it does not stop anyone moving forward.

_It is a floor, not a wall._

PR #1857, opened 2026-08-28
Pixi: Add colorama dependency

penww:
## Description

Fix CI issue for https://github.com/ros2/ros2cli/pull/1248. In that PR, we added color support for ros2cli, `python3-colorama` is used for cross-platform colored terminal output.

### Did you use Generative AI?

No.

### Additional Information

- Related fix PR: https://github.com/ros2/ci/pull/892

PR #1860, opened 2026-08-29
Add a pixi.lock

mjcarroll:
`pixi.toml` has been here since the pixi environment landed, but the lock file it produces never was, so nothing recorded which packages a given manifest actually resolved to, two people running `pixi install` a week apart could get different environments.

Written by pixi 0.78 in lock format version 7, covering both platforms the manifest currently declares.

`.gitattributes` already marks `pixi.lock` as `merge=binary` and `linguist-generated`, so it stays collapsed in diffs.

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -1,0 +1,8822 @@
+version: 7
+platforms:
+- name: linux-64
+  virtual-packages:
+  - __unix=0=0
+  - __linux=4.18
+  - __glibc=2.28
+  - __archspec=0=x86_64
+- name: win-64
+  virtual-packages:
+  - __win=10.0
+  - __archspec=0=x86_64
+environments:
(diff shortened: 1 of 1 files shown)


christophebedard:
Can we use GitHub stacked PRs here?

mjcarroll:
We could except I did it from a fork and I guess that doesn't work?

christophebedard:
that would be annoying

christophebedard (approved):
LGTM.

I'll let @knmcguire review too given the current discussion on Zulip

mjcarroll:
Updated to include #1857

mjcarroll:
Lock file itself is benign, I'm going to merge with the single review, but would appreciate @knmcguire input on the next one.

PR #1861, opened 2026-08-29
Check pixi.lock on every pull request

mjcarroll:
A change to `pixi.toml` invalidates the `pixi.lock` next to it. This refuses the merge unless both are satisfied: the manifest solves on every platform it declares, and the lock file in the pull request is the one that solve produces.

`pixi lock --check` does both halves in one command.  It solves each platform under `workspace.platforms` and exits non-zero when a platform has no solution *or* when the result differs from the committed lock.  When the lock is just stale, the corrected file is uploaded as an artifact, so committing it does not require installing pixi first.

After merging, I will mark **`Pixi Lock / Pixi lock is valid`** as a required status check

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,5 @@
 # SCM syntax highlighting & preventing 3-way merges
 pixi.lock merge=binary linguist-language=YAML linguist-generated=true
+
+# Shell scripts must keep LF endings to run on Linux runners.
+*.sh text eol=lf
--- a/.github/scripts/pixi-lock-changed-files.sh
+++ b/.github/scripts/pixi-lock-changed-files.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+# Sets `pixi=true|false` on ${GITHUB_OUTPUT} depending on whether the pull
+# request touches the pixi files.  The workflow has no `paths:` filter because a
+# filtered-out required check never reports, so it decides here instead.
+#
+# Environment: GH_TOKEN, GITHUB_REPOSITORY, PR_NUMBER.
+
+set -euo pipefail
+
+: "${GITHUB_REPOSITORY:?must be set}"
+: "${PR_NUMBER:?must be set}"
+
+# Unset outside Actions.
+GITHUB_OUTPUT="${GITHUB_OUTPUT:-/dev/null}"
+GITHUB_STEP_SUMMARY="${GITHUB_STEP_SUMMARY:-/dev/null}"
+
+watched='pixi\.(toml|lock)|\.github/workflows/pixi-lock\.yaml|\.github/scripts/pixi-lock-[a-z-]+\.sh'
+
+changed="$(gh api "repos/${GITHUB_REPOSITORY}/pulls/${PR_NUMBER}/files" \
+  --paginate --jq '.[].filename')"
+
+if grep -qxE "${watched}" <<< "${changed}"; then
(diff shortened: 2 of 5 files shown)


mjcarroll:
Open questions (from @knmcguire), some of this answered via: https://pixi.prefix.dev/latest/workspace/lock_file/#committing-your-lock-file

1. When does this get updated?

I would expect that something like renovate bot does it on a regular cadence (once a week seems reasonable)?

2. Should we turn on `--frozen` and `--locked` in CI to enforce this?

Probably?  Would make sense to at least try

3. What about other PRs that touch pixi.toml but want to run through CI?

We can probably turn off `--frozen` and `--locked` in the case that a user provides another toml file.

For now, the only enforcement would be that the pixi.toml file resolves to a valid set of dependencies and generates a pixi.lock file and that pixi.lock file is committed to `ros2/ros2`.  CI, etc can be considered follow-on work here, as this will just provide a store of the resolved dependencies over time.

knmcguire:
Just to add to 2. we already use --frozen for the run in CI already (at least for windows):

https://github.com/ros2/ci/blob/0c8d1b1c5ab6f8f6f6d0b0210342311b382d1d1d/windows_docker_resources/Dockerfile#L76-L77

What I perhaps not exactly understand is that they say that you do --frozen OR --locked 🤔 so not exactly sure what the implications are. Will it simply fail if we do pixi install --frozen?

knmcguire:
Alright, did some research on this, and for CI they do recommend pixi install --locked and pixi run --locked. They will both fail if the .lock file is out of sync with the manifest file (toml file), while --frozen will just execute the lock file as is, disregarding if anything needs updating. (https://pixi.prefix.dev/latest/reference/cli/pixi/install/#update-options)

BUT, if anything, and I mean anything, both the packages or the dependencies they rely on, have an updated version, this will make the CI fail when we use --locked anywhere. This doesn't only happen with pixi.toml updates, but this is an continuous threat where the state can change daily and with every windows ci run. 

So what I would perhaps propose, is to test here in ci a pixi install --locked (update: I said 'frozen' before), or using what prefix themselve proposes here: https://pixi.prefix.dev/latest/integration/ci/github_actions/#environment-activation (with locked on true and pixi install right underneath). That should already do what you want to as that will failed with any unsynced .toml file. 

For in the CI let's keep it using --frozen for both the install and run, as I think it will be a little too disruptive.

mjcarroll:
> This doesn't only happen with pixi.toml updates, but this is an continuous threat where the state can change daily and with every windows ci run.

Should we just use `--frozen` in CI then?  That would make it use the lock file exactly as committed, with something else regularly updating the lock file (maybe a github action/rennovate?)

I don't want CI to break just because a patch version of a dependency changes.

knmcguire (approved):
Yeah indeed. Let's keep using --frozen (we will also need to add it [here](https://github.com/ros2/ci/blob/d24bc438376706eacacab0d84a9034e60bafa41e/windows_docker_resources/Dockerfile#L58))

About the github action, I think it looks good! There is a lot of scripting happening in the github action itself that maybe could come from from somewhere else, but for a first trial I would say that this is good. At least it describes clearly what needs to be done if something fails and I like it that it actually provides the actual pixi.toml file necessary for the PR to pass

knmcguire:
Ah yes much better. At least it is possible to adjust the scripts easier by hand

mjcarroll:
I went ahead and added shellcheck here as well.

PR #1872, opened 2026-09-12
Add a pixi.lock (backport #1860)

dskkato:
`pixi.toml` has been here since the pixi environment landed, but the lock file it produces never was, so nothing recorded which packages a given manifest actually resolved to, two people running `pixi install` a week apart could get different environments.

Written by pixi 0.78 in lock format version 7, covering both platforms the manifest currently declares.

`.gitattributes` already marks `pixi.lock` as `merge=binary` and `linguist-generated`, so it stays collapsed in diffs.

----

This is a manual backport of #1860, but generated from the lyrical's pixi.toml. 

Note that this PR is prepared to enable semi-automatic backport of #1861 in the future, if needed.

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -1,0 +1,8730 @@
+version: 7
+platforms:
+- name: linux-64
+  virtual-packages:
+  - __unix=0=0
+  - __linux=4.18
+  - __glibc=2.28
+  - __archspec=0=x86_64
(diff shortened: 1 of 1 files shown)


christophebedard:
@mjcarroll did we want to also add a `pixi.lock` file on `lyrical`?

PR #1875, opened 2026-09-18
Update lyrical ros2.repos for 2026-09-18 sync

sloretz:
## Description

Updating released versions for today's sync

### Did you use Generative AI?

no

### Additional Information

https://discourse.openrobotics.org/t/preparing-for-lyrical-sync-2026-09-17/58150/5


christophebedard:
aren't we only doing this for patch releases, not syncs?

sloretz:
> aren't we only doing this for patch releases, not syncs?

I wasn't sure. It's definitely only necessary for patch releases, but I figured there was no harm in updating it after a sync.

